PR #1, opened 2016-11-09
Add build support files

ankurcha:
* Add `.gitignore` file to make sure unintended files are not added.
* Add `.travis.yaml` to do automated CI builds.
* Add `Dockerfile` to create docker image (very useful when deploying to kuberentes)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,8 +1,6 @@
 FROM maven:3-jdk-8-onbuild
 
-ENV JVM_GC_OPTS="-XX:+UseG1GC -XX:MaxGCPauseMillis=200 -XX:+UseStringDeduplication -XX:+ParallelRefProcEnabled"
-COPY docker-entrypoint.bash /
-RUN chmod a+x /docker-entrypoint.bash
+ENV JVM_OPTS=""
 EXPOSE 9411
 
-ENTRYPOINT ["/docker-entrypoint.bash"]
+ENTRYPOINT [ "sh", "-c", "java $JVM_OPTS -Djava.security.egd=file:/dev/./urandom -jar /usr/src/app/server/target/server-*.jar" ]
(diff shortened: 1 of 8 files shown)


ankurcha:
I signed it!

Dockerfile
```diff
@@ -0,0 +1,8 @@
+FROM maven:3-jdk-8-onbuild
+
+ENV JVM_GC_OPTS="-XX:+UseG1GC -XX:MaxGCPauseMillis=200 -XX:+UseStringDeduplication -XX:+ParallelRefProcEnabled"
```
ankurcha: The idea behind putting an entry point is to make sure that we are able to put some additional Java opts.

Does that make sense? I used the pattern from the appengine-openjdk runtime docker image.

kevinmdavis: I think the user could set the JAVA_OPTS environment variable and the Dockerfile could pass that directly to the java command.
```
CMD java $JAVA_OPTS -jar /usr/src/app/server/target/server-*.jar
```
and then the user can run the container by:
```
docker run -e JAVA_OPTS="-Xmx1024M -XX:+UseG1GC" "stackdriver-zipkin-server"
```

I think in this case, we don't need a separate entrypoint, and the user can pass in their JVM args. Would this approach work for your use case?

ankurcha:
I updated the docker file as per the code review feedback.

kevinmdavis:
Thanks! Can we add the Kubernetes docs as a separate pull request? I can go ahead and merge your other changes, and I'll work on getting the container published. After that's done, we can update the Kubernetes docs with the official container location.

ankurcha:
I'll remove kube from here and make a new pr. Thanks!

ankurcha:
Done! Should be good to go.

PR #4, opened 2016-11-14
Add kubernetes deployment manifest

ankurcha:
* Added a kube deployment manifest.


kube-manifest.sample.yaml
```diff
@@ -0,0 +1,39 @@
...
+    spec:
+      replicas: 1
+      template:
+        metadata:
+          labels:
+            app: zipkin-stackdriver
+        spec:
+          containers:
+            - name: zipkin-stackdriver
+              image: GoogleCloudPlatform/stackdriver-zipkin:latest
```
kevinmdavis: I've hosted the docker image at b.gcr.io/stackdriver-trace-docker/zipkin-collector. We can reference that here.

kube-manifest.sample.yaml
```diff
@@ -0,0 +1,39 @@
...
+              imagePullPolicy: Always
+              env:
+                - name: PROJECT_ID
+                  value: my_project_id
+                - name: STORAGE_TYPE
+                  value: stackdriver
```
kevinmdavis: This is the default value for our zipkin collector, so this isn't required.

ankurcha: updated

kube-manifest.sample.yaml
ankurcha: updated

PR #10, opened 2016-12-19
Updates to latest zipkin and switches to 128-bit trace IDs

adriancole:
This changes to the latest version of Zipkin and adjusts the code to use
128-bit trace ids as defined by Zipkin. This fills high bits with zero
when the trace doesn't have high bits set. This would occur when 64-bit
trace IDs are in use.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -41,6 +41,10 @@
     <tag>HEAD</tag>
   </scm>
 
+  <properties>
+    <zipkin.version>1.18.0</zipkin.version>
+  </properties>
+
   <build>
     <pluginManagement>
       <plugins>
--- a/translation/src/main/java/com/google/cloud/trace/zipkin/translation/TraceTranslator.java
+++ b/translation/src/main/java/com/google/cloud/trace/zipkin/translation/TraceTranslator.java
@@ -55,7 +55,7 @@ public class TraceTranslator {
 
     Collection<Trace> translatedTraces = new ArrayList<>();
     for (Span zipkinSpan : groupedSpans) {
-      String traceId = convertTraceId(zipkinSpan.traceId);
+      String traceId = convertTraceId(zipkinSpan);
       if (currentTrace == null || !traceId.equals(currentTrace.getTraceId())) {
         finishTrace(currentTrace, translatedTraces);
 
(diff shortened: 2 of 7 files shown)


translation/src/main/java/com/google/cloud/trace/zipkin/translation/TraceTranslator.java
```diff
@@ -89,15 +89,13 @@ private void finishTrace(Trace.Builder traceBuilder, Collection<Trace> converted
...
-    for (byte b : idBuffer.array()) {
-      idBuilder.append(String.format("%02x", b));
-    }
-    return idBuilder.toString();
+  // Stackdriver trace ID's are 128 bits = 16 bytes * 8
+  private String convertTraceId(Span zipkinSpan) {
+    // zipkin conditionally encodes its id as 16 chars if the high bits are zero
+    // This unconditionally encodes as 32 chars as Stackdriver trace IDs are fixed length.
+    char[] result = new char[32];
+    Util.writeHexLong(result, 0, zipkinSpan.traceIdHigh);
```
adriancole: ps if you don't like this, I can switch back to the other way, np

kevinmdavis: I'm OK with this if you're OK with this depending on zipkin.internal.Util.

We may want to have this write `zipkinSpan.traceId` for both the high and low bits if `zipkinSpan.traceIdHigh` isn't set. Some users call our API and order results by traceId to get a "random sample" of traces. Duplicating the 64-bit ID instead of padding the ID with zeros will remove some sampling bias for any projects that contain both Zipkin and Stackdriver traces. Do you think that would cause any problems?

pom.xml
```diff
@@ -41,6 +41,10 @@
     <tag>HEAD</tag>
   </scm>
 
+  <properties>
+    <zipkin.version>1.18.0</zipkin.version>
```
kevinmdavis: Thanks!

translation/src/main/java/com/google/cloud/trace/zipkin/translation/TraceTranslator.java
adriancole: the only problem with duplicating or otherwise is that the id won't match what's put in the logging context. Interesting use case on the "random sample" this will clash on that, though curious why they don't shuffle instead? Is this tool OSS (ex can we open an issue to discuss that?)

adriancole: incidentally, we can always change this later.. I'll do this.. will revert to the original impl, but add a comment on why we are duplicating. Later it can be followed-up as necessary.

adriancole:
PTAL I think all good now?

translation/src/main/java/com/google/cloud/trace/zipkin/translation/TraceTranslator.java
kevinmdavis: No, it's not an OSS tool. However, it is a pretty small use case, and I'm not sure if there are many projects that will have a mix of Stackdriver and Zipkin traces. The mismatch with the logging context may be a bigger issue if you think Zipkin users reference that frequently. We can pad with 0's if you think that would be better. Anyway, this problem won't exist once Zipkin trace ID's are 128-bit. :)

kevinmdavis: However, if you do pad with 0's, TraceTranslatorTest:testTranslateTrace needs to be updated.

adriancole:
made some notes on the next commit. PS most major tracers are 128-bit now,
just that they don't issue them by default
https://github.com/openzipkin/b3-propagation/issues/6

adriancole:
padding 0s left would be better I think.. it switches the explanation from
why the trace id doesn't match to why lex sorting on trace id doesn't end
up with random sample.

adriancole:
okie split into two commits.. lemme know if you want one (or squash). ttfn!

PR #13, opened 2016-12-20
Added env variables to Docker runner

denyska:
Made README.md more explicit: it was bit unclear how to run adapter within docker w/o credentials.

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,13 +26,21 @@ additional configuration is required.
 ```
 java -jar collector.jar
 ```
+or just
+```
+./collector.jar
+```
+
+If docker is used from a GCE host, authentication will happen automatically and Zipkin collector can be started with:
+```
+docker run -p 9411:9411 b.gcr.io/stackdriver-trace-docker/zipkin-collector
+```
+
 
 #### Using an explicit projectId and credentials file path
 ```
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -34,7 +34,7 @@ GOOGLE_APPLICATION_CREDENTIALS="/path/to/credentials.json" PROJECT_ID="my_projec
 
 #### Using Docker
 ```
-docker run -p 9411:9411 b.gcr.io/stackdriver-trace-docker/zipkin-collector
+docker run -v /path/to_credentials:/opt/gcloud -e GOOGLE_APPLICATION_CREDENTIALS="/opt/gcloud/credentials.json" -e PROJECT_ID="my_project_id" -p 9411:9411 b.gcr.io/stackdriver-trace-docker/zipkin-collector
```
kevinmdavis: I think this new example will be good to have, but users on GCE could still use the existing example. Do you think it would make sense to move this example underneath the "Running on GCE" and "Using an explicit projectId and credentials file path" sections? Each of those sections could have an example running the JAR and an example using docker.

denyska: @kevinmdavis addressed comments, please take a look.

PR #15, opened 2016-12-22
Made executor service configurable

denyska:
Before the change com.google.cloud.trace.zipkin.StackdriverStorageComponent was essentially single-threaded as ThreadPoolTaskExecutor it was using was configured with default values: corePoolSize = 1, maxPoolSize = Integer.MAX_VALUE, queueCapacity = Integer.MAX_VALUE 

From JAVA doc of java.util.concurrent.ThreadPoolExecutor

>  * When a new task is submitted in method {@link #execute(Runnable)},
>  * and fewer than corePoolSize threads are running, a new thread is
>  * created to handle the request, even if other worker threads are
>  * idle.  If there are more than corePoolSize but less than
>  * maximumPoolSize threads running, a new thread will be created only
>  * if the queue is full. 

That means unless queue reaches Integer.MAX_VALUE  no new threads would be allocated.

### DIFF
--- a/autoconfigure-storage/src/main/java/com/google/cloud/trace/zipkin/autoconfigure/ZipkinStackdriverStorageAutoConfiguration.java
+++ b/autoconfigure-storage/src/main/java/com/google/cloud/trace/zipkin/autoconfigure/ZipkinStackdriverStorageAutoConfiguration.java
@@ -21,12 +21,8 @@ import com.google.cloud.trace.grpc.v1.GrpcTraceSink;
 import com.google.cloud.trace.v1.sink.TraceSink;
 import com.google.cloud.trace.zipkin.StackdriverStorageComponent;
 import com.google.common.io.ByteStreams;
-import java.io.IOException;
-import java.io.InputStream;
-import java.net.HttpURLConnection;
-import java.net.URL;
-import java.util.Collections;
-import java.util.concurrent.Executor;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.beans.factory.annotation.Qualifier;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnMissingBean;
@@ -37,6 +33,13 @@ import org.springframework.context.annotation.Configuration;
 import org.springframework.scheduling.concurrent.ThreadPoolTaskExecutor;
 import zipkin.storage.StorageComponent;
 
+import java.io.IOException;
+import java.io.InputStream;
+import java.net.HttpURLConnection;
+import java.net.URL;
+import java.util.Collections;
+import java.util.concurrent.Executor;
+
 /**
  * Auto-configuration to set the Stackdriver StorageComponent as the Zipkin storage backend.
  */
@@ -45,14 +48,21 @@ import zipkin.storage.StorageComponent;
 @ConditionalOnProperty(name = "zipkin.storage.type", havingValue = "stackdriver")
 @ConditionalOnMissingBean(StorageComponent.class)
 public class ZipkinStackdriverStorageAutoConfiguration {
+  private static final Logger log = LoggerFactory.getLogger(ZipkinStackdriverStorageAutoConfiguration.class);
+
   @Autowired
   ZipkinStackdriverStorageProperties storageProperties;
 
-  @Bean
+  @Bean(name = "stackdriverExecutor")
   @ConditionalOnMissingBean(Executor.class) Executor executor() {
     ThreadPoolTaskExecutor executor = new ThreadPoolTaskExecutor();
     executor.setThreadNamePrefix("ZipkinStackdriverStorage-");
+    executor.setCorePoolSize(storageProperties.getExecutor().getCorePoolSize());
+    executor.setMaxPoolSize(storageProperties.getExecutor().getMaxPoolSize());
+    executor.setQueueCapacity(storageProperties.getExecutor().getQueueCapacity());
     executor.initialize();
+
+    log.info("Configured Executor for ZipkinStackDriver Storage with: {}", storageProperties.getExecutor());
     return executor;
   }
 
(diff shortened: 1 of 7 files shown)


autoconfigure-storage/src/main/java/com/google/cloud/trace/zipkin/autoconfigure/ZipkinStackdriverStorageAutoConfiguration.java
```diff
@@ -44,15 +46,21 @@
 @EnableConfigurationProperties(ZipkinStackdriverStorageProperties.class)
 @ConditionalOnProperty(name = "zipkin.storage.type", havingValue = "stackdriver")
 @ConditionalOnMissingBean(StorageComponent.class)
+@Slf4j
 public class ZipkinStackdriverStorageAutoConfiguration {
   @Autowired
   ZipkinStackdriverStorageProperties storageProperties;
 
-  @Bean
+  @Bean(name = "stackDriverExecutor")
```
kevinmdavis: stackDriver -> stackdriver

autoconfigure-storage/src/main/java/com/google/cloud/trace/zipkin/autoconfigure/ZipkinStackdriverStorageProperties.java
```diff
@@ -16,26 +16,36 @@
...
 
+import lombok.Data;
 import org.springframework.boot.context.properties.ConfigurationProperties;
 
 @ConfigurationProperties("zipkin.storage.stackdriver")
+@Data
 public class ZipkinStackdriverStorageProperties {
   private String projectId;
   private String apiHost = "cloudtrace.googleapis.com";
+  private Executor executor = new Executor();
```
kevinmdavis: How does this work with the configuration yml file? Can the user set zipkin.storage.stackdriver.executor.corePoolSize?

storage/src/test/java/com/google/cloud/trace/zipkin/StackdriverStorageComponentTest.java
```diff
@@ -0,0 +1,117 @@
...
+  }
+
+  private TraceSink createSlowTraceSink(final long grpcDelayInMillis)
+  {
+    return new TraceSink()
+    {
+      @Override
+      public void receive(Trace trace)
+      {
+        //System.out.println(format("Processing trace %s on thread: %s", trace.getTraceId(), Thread.currentThread().getName()));
```
kevinmdavis: Delete this comment.

autoconfigure-storage/src/main/java/com/google/cloud/trace/zipkin/autoconfigure/ZipkinStackdriverStorageProperties.java
```diff
@@ -16,26 +16,36 @@
 
 package com.google.cloud.trace.zipkin.autoconfigure;
 
+import lombok.Data;
 import org.springframework.boot.context.properties.ConfigurationProperties;
 
 @ConfigurationProperties("zipkin.storage.stackdriver")
+@Data
```
adriancole: ps lombok has proven tedious to support, hence removed in spring cloud etc.

autoconfigure-storage/src/main/java/com/google/cloud/trace/zipkin/autoconfigure/ZipkinStackdriverStorageProperties.java
adriancole: here's an example of how to test this

https://github.com/openzipkin/zipkin/pull/1450/files#diff-d5ed7e4d8a81b9885985b99989e2a628R93

conventionally, snake case is preferred with properties.

the main server readme has a bit on how to set things like this (which is a summary of spring boot docs)
https://github.com/openzipkin/zipkin/blob/master/zipkin-server/README.md#configuration-for-the-ui

denyska: yes, for example config below will honored 

```
zipkin:
  storage:
    stackdriver:
      executor:
        corePoolSize: 100
        maxPoolSize: 5000
```

autoconfigure-storage/src/main/java/com/google/cloud/trace/zipkin/autoconfigure/ZipkinStackdriverStorageAutoConfiguration.java
denyska: addressed

storage/src/test/java/com/google/cloud/trace/zipkin/StackdriverStorageComponentTest.java
denyska: addressed

autoconfigure-storage/src/main/java/com/google/cloud/trace/zipkin/autoconfigure/ZipkinStackdriverStorageProperties.java
denyska: done

denyska:
@kevinmdavis @adriancole updated PR based on comments, squashed all commits into one so it looks neat and tidy

autoconfigure-storage/src/main/java/com/google/cloud/trace/zipkin/autoconfigure/ZipkinStackdriverStorageAutoConfiguration.java
```diff
@@ -45,14 +48,21 @@
...
+  @Bean(name = "stackdriverExecutor")
   @ConditionalOnMissingBean(Executor.class) Executor executor() {
     ThreadPoolTaskExecutor executor = new ThreadPoolTaskExecutor();
     executor.setThreadNamePrefix("ZipkinStackdriverStorage-");
+    executor.setCorePoolSize(storageProperties.getExecutor().getCorePoolSize());
+    executor.setMaxPoolSize(storageProperties.getExecutor().getMaxPoolSize());
+    executor.setQueueCapacity(storageProperties.getExecutor().getQueueCapacity());
     executor.initialize();
+
+    log.info("Configured Executor for ZipkinStackDriver Storage with: {}", storageProperties.getExecutor());
```
adriancole: nit I'd use jul since this is not advanced and it generally dodges a topic of which logging system to use.

storage/pom.xml
```diff
@@ -44,6 +44,23 @@
...
+    </dependency>
+    <dependency>
+      <groupId>commons-logging</groupId>
+      <artifactId>commons-logging</artifactId>
+      <version>1.1.1</version>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>pl.pragmatists</groupId>
+      <artifactId>JUnitParams</artifactId>
```
adriancole: personally prefer assertj, and it is used in all zipkin java projects, but by no means required. just sayin

Issue #23, opened 2017-01-18
Add download link


saturnism:
There is no download link of the distribution. The package in GitHub releases is the source.

Recommend adding a Download section and refer to all release versions to this link:
https://mvnrepository.com/artifact/com.google.cloud.trace.adapters.zipkin/collector

adriancole:
we can also add a maven central badge
https://github.com/jirutka/maven-badges

denyska:
somehow release process has to be synchronized too: latest docker image doesn't conform with maven-central.

kevinmdavis:
* I've previously tagged the `gcr.io/stackdriver-trace-docker/zipkin-collector:v0.2.0` docker image.
* I've just pushed the v0.2.0 release to Maven central so it be visible soon.

kevinmdavis:
I've updated the README to include this information.

PR #25, opened 2017-01-20
Update README

kevinmdavis:
#23

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,12 +17,18 @@ A drop-in replacement for the standard Zipkin HTTP collector that writes to the
 Stackdriver Trace service.
 
 ### Configuration
+
 |Environment Variable           | Value            |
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -17,12 +17,20 @@ A drop-in replacement for the standard Zipkin HTTP collector that writes to the
 Stackdriver Trace service.
 
 ### Configuration
+Typically, sampling is done by the Tracer prior to sending data to the Collector, so the Collector
```
adriancole: shouldn't this be in the table as a description of COLLECTOR_SAMPLE_RATE? or maybe a section called sampling?

kevinmdavis: Yeah, that probably makes more sense. I had forgotten zipkin-server exposed this as an environment variable.

PR #31, opened 2017-03-14
Force GRPC/Netty upgrade 

denyska:
Force GRPC upgrade to pickup https://github.com/netty/netty/commit/94d7557dead5e7780769d44bbb4aa7530149ca34 that supposedly fixes memory leak.


adriancole:
LGTM

PR #32, opened 2017-03-24
Refactored Zipkin Proxy artifacts

saturnism:
- moved EnableStackDriver.* to the `autoconfigure` module
- created a starter `pom.xml` so that others can build their own collector (and override things like `Credentials` easily)
- changed the resolution of `projectId` property


starter/pom.xml
```diff
@@ -0,0 +1,62 @@
...
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+
+    <parent>
+        <groupId>com.google.cloud.trace.adapters.zipkin</groupId>
+        <artifactId>parent</artifactId>
+        <version>0.2.1-SNAPSHOT</version>
+    </parent>
+
+    <artifactId>starter</artifactId>
```
adriancole: ps it has been my experience that making ad-hoc servers ends up as a support burden, as there's rarely a reason to create a custom server.  starters are usually a means to create custom servers... are you sure you want to do this?

currently, we are using the "module" approach when creating custom servers, so it still ends up something that needs no custom build, rather composition at runtime.. like this: https://github.com/openzipkin/zipkin-aws/tree/master/autoconfigure/collector-sqs#running

If we want to bring the tomcat changes needed upstream to support that, seems great to me..

adriancole: ps by support burden I meant people coming into gitter who found the starter and started playing with it but don't know anything about tracing or what they need it for. requests for support on how to get eclipse working because they feel they need a custom server to start zipkin.. or blogs that explain how to do basic things already done by the default server, which propagate misunderstanding on how to configure things or that basic things don't work out-of-box. things like this

PR #41, opened 2017-08-02
Add docs for COLLECTOR_SAMPLE_RATE allowed values

benhass:
The existing doc is somewhat misleading. If you use COLLECTOR_SAMPLE_RATE=100 the collector will fail with an exception:

```
java.lang.IllegalArgumentException: rate should be between 0 and 1: was 100.0
```


benhass:
I signed it

PR #47, opened 2017-10-24
Migrate business logic to use native zipkin2 API.

anuraaga:
Apologies for the formatting changes - I was finding style inconsistencies and having trouble writing the code without a consistent style definition, so I put everything through google-java-format which I believe is the standard for GoogleCloudPlatform projects. If it's too much trouble to include reformatting here, I can try creating a PR with just the logic changes, but it'll be quite some IDE fighting...

### DIFF
--- a/autoconfigure-storage/src/main/java/com/google/cloud/trace/zipkin/autoconfigure/ZipkinStackdriverStorageAutoConfiguration.java
+++ b/autoconfigure-storage/src/main/java/com/google/cloud/trace/zipkin/autoconfigure/ZipkinStackdriverStorageAutoConfiguration.java
@@ -24,44 +24,38 @@ import com.google.cloud.trace.zipkin.StackdriverStorageComponent;
 import com.google.common.base.Preconditions;
 import com.google.common.io.ByteStreams;
 import io.netty.handler.ssl.OpenSsl;
+import java.io.IOException;
+import java.io.InputStream;
+import java.net.HttpURLConnection;
+import java.net.URL;
+import java.util.Collections;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.beans.factory.annotation.Qualifier;
-import org.springframework.boot.actuate.endpoint.PublicMetrics;
-import org.springframework.boot.actuate.metrics.Metric;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnMissingBean;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty;
 import org.springframework.boot.context.properties.EnableConfigurationProperties;
 import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.Configuration;
 import org.springframework.scheduling.concurrent.ThreadPoolTaskExecutor;
-import zipkin.storage.StorageComponent;
+import zipkin.internal.V2StorageComponent;
+import zipkin2.storage.StorageComponent;
 
-import java.io.IOException;
-import java.io.InputStream;
-import java.net.HttpURLConnection;
-import java.net.URL;
-import java.util.ArrayList;
-import java.util.Collection;
-import java.util.Collections;
-import java.util.concurrent.Executor;
-
-/**
- * Auto-configuration to set the Stackdriver StorageComponent as the Zipkin storage backend.
- */
+/** Auto-configuration to set the Stackdriver StorageComponent as the Zipkin storage backend. */
 @Configuration
 @EnableConfigurationProperties(ZipkinStackdriverStorageProperties.class)
 @ConditionalOnProperty(name = "zipkin.storage.type", havingValue = "stackdriver")
 @ConditionalOnMissingBean(StorageComponent.class)
 public class ZipkinStackdriverStorageAutoConfiguration {
-  private static final Logger log = LoggerFactory.getLogger(ZipkinStackdriverStorageAutoConfiguration.class);
+  private static final Logger log =
+      LoggerFactory.getLogger(ZipkinStackdriverStorageAutoConfiguration.class);
 
-  @Autowired
-  ZipkinStackdriverStorageProperties storageProperties;
+  @Autowired ZipkinStackdriverStorageProperties storageProperties;
 
   @Bean(name = "stackdriverExecutor")
-  @ConditionalOnMissingBean(ThreadPoolTaskExecutor.class) ThreadPoolTaskExecutor executor() {
+  @ConditionalOnMissingBean(ThreadPoolTaskExecutor.class)
+  ThreadPoolTaskExecutor executor() {
     ThreadPoolTaskExecutor executor = new ThreadPoolTaskExecutor();
     executor.setThreadNamePrefix("ZipkinStackdriverStorage-");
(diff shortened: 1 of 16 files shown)


anuraaga:
@bogdandrutu @adriancole PTAL

translation/src/test/java/com/google/cloud/trace/zipkin/translation/TraceTranslatorTest.java
```diff
@@ -112,27 +73,51 @@ public void testTranslateTrace() {
...
-        .addAnnotation(Annotation.create(1474488797500000L, SERVER_RECV, serverEp))
-        .addAnnotation(Annotation.create(1474488798300000L, SERVER_SEND, serverEp))
-        .build();
-    Span span4 = Span.builder().traceId(1).parentId(2L).id(3).name("custom-span")
-        .timestamp(1474488797600000L)
-        .duration(200000L)
-        .build();
+    Span span1 =
+        Span.newBuilder()
+            // TODO(anuraag): This seems like it should have KIND == SERVER, but that causes the
```
anuraaga: Please check this note that is left over, I'm not sure if my reasoning is wrong, or the parent ID rewriting logic is wrong.

adriancole:
>
>  If it's too much trouble to include reformatting here, I can try creating
> a PR with just the logic changes, but it'll be quite some IDE fighting...
>
sob story but once I had to do java -jar
~/Downloads/google-java-format-1.4-all-deps.jar to get things right due to
import order bug w/in intellij. seems might be fixed in 1.5 now but not
verified
https://github.com/google/google-java-format/commit/adf8837b70dbc03bc25679aa54dc6d8229c50861

autoconfigure-storage/src/main/java/com/google/cloud/trace/zipkin/autoconfigure/ZipkinStackdriverStorageAutoConfiguration.java
```diff
@@ -106,15 +101,22 @@ String getDefaultProjectId() throws IOException {
     HttpURLConnection connection = (HttpURLConnection) url.openConnection();
     connection.setRequestProperty("Metadata-Flavor", "Google");
     connection.setRequestMethod("GET");
-    try (InputStream responseStream = connection.getInputStream())
-    {
+    try (InputStream responseStream = connection.getInputStream()) {
```
adriancole: fyi your formatting changes look legit to me... although these changes move it to google style from what is some variation of that or square.. later it will switch to square :P

storage/src/main/java/com/google/cloud/trace/zipkin/StackdriverSpanConsumer.java
```diff
@@ -33,15 +43,81 @@
...
   }
 
   @Override
-  public void accept(List<Span> spans) {
-    Collection<Trace> traces = translator.translateSpans(spans);
-    consumer.receive(Traces.newBuilder().addAllTraces(traces).build());
+  public Call<Void> accept(List<Span> spans) {
+    final Traces traces = Traces.newBuilder().addAllTraces(translator.translateSpans(spans)).build();
+    try {
+      ListenableFuture<Void> future =
```
adriancole: it might be slightly more direct to not use ListenableFuture here as for example call.execute can just call `consumer.receive(traces);` directly.  I'd probably do it that way, extending from base call (which we ought to make a public type :P )

in any case, the point of returning call is to return something not started, so I'd move it if you can.

anuraaga: Got it - a little annoying duplicating all the state from the `Future` but went ahead and moved the execution into the `Call`, PTAL

bogdandrutu:
Please rebase the PR to fix the conflicts.

anuraaga:
Merged

storage/src/main/java/com/google/cloud/trace/zipkin/StackdriverSpanConsumer.java
```diff
@@ -33,15 +38,93 @@
...
+        Traces.newBuilder().addAllTraces(translator.translateSpans(spans)).build();
+    return new TracesCall(traces);
+  }
+
+  private class TracesCall extends Call<Void> {
+
+    private final Traces traces;
+
+    private volatile boolean canceled;
+    private AtomicBoolean executed;
```
adriancole: nit: final

storage/src/main/java/com/google/cloud/trace/zipkin/StackdriverSpanConsumer.java
```diff
@@ -33,15 +38,93 @@
...
+
+    @Override
+    public void enqueue(Callback<Void> callback) {
+      markExecuted();
+      if (isCanceled()) {
+        callback.onError(new IOException("Canceled"));
+        return;
+      }
+      try {
+        future =
```
adriancole: I wouldn't bother with making a future here as you can't really cancel it. probably simpler to just execute a runnable right?

anuraaga: Cool - simplified the cancellation logic.

anuraaga:
@adriancole Do you mind clarifying this point I was confused about?

https://github.com/GoogleCloudPlatform/stackdriver-zipkin/pull/47/files#r146511723

Also, I noticed I am using a single internal class, zipkin.internal.V2StorageComponent - do you think I should shade that?

translation/src/test/java/com/google/cloud/trace/zipkin/translation/TraceTranslatorTest.java
adriancole: definitely if a span ends up with "sr" "ss" something is wrong :)

autoconfigure-storage/src/main/java/com/google/cloud/trace/zipkin/autoconfigure/ZipkinStackdriverStorageAutoConfiguration.java
```diff
@@ -24,52 +24,48 @@
...
-
-import java.io.IOException;
-import java.io.InputStream;
-import java.net.HttpURLConnection;
-import java.net.URL;
-import java.util.ArrayList;
-import java.util.Collection;
-import java.util.Collections;
-import java.util.concurrent.Executor;
+import zipkin.internal.V2StorageComponent;
```
adriancole: this part is ok as it is tightly coupled to the server (autoconfig) anyway. don't shade anything here.

translation/src/test/java/com/google/cloud/trace/zipkin/translation/TraceTranslatorTest.java
anuraaga: Ah ok - I went ahead and removed all the sr/cr type annotations from the tests to make that clearer.

bogdandrutu:
Please rebase and fix all the comments.

adriancole:
afaict there's no outstanding comments (as shade is actually not used anymore). all left is fix conflicts. This will help in refactoring the project later to be normal zipkin plugin.

anuraaga:
Rebased - thanks

anuraaga:
@bogdandrutu friendly ping

adriancole:
umm... folks? If you want to give me access now I can merge this :) Otherwise, can we figure out what's up?

PR #63, opened 2018-03-21
Reinstates translation module

adriancole:
It seems there were external users of the trace translator, for example:
https://github.com/curioswitch/curiostack/blob/master/common/server/framework/src/main/java/org/curioswitch/common/server/framework/monitoring/StackdriverReporter.java

This re-instates the translation module, which also allows us to
decouple from a hard dependency in the future.


storage-stackdriver/pom.xml
```diff
@@ -31,15 +31,8 @@
...
-      <artifactId>grpc-google-devtools-cloudtrace-v1</artifactId>
-      <version>${grpc-google-devtools-cloudtrace-v1.version}</version>
-      <exclusions>
-        <exclusion>
-          <groupId>io.grpc</groupId>
-          <artifactId>grpc-all</artifactId>
-        </exclusion>
-      </exclusions>
+      <groupId>${project.groupId}</groupId>
+      <artifactId>zipkin-translation-stackdriver</artifactId>
```
adriancole: note: later we will scrub this dep in the collector as it will eventually become problematic when we do multiple stackdriver versions.

translation-stackdriver/src/main/java/zipkin2/translation/stackdriver/SpanTranslator.java
```diff
@@ -11,15 +11,13 @@
  * or implied. See the License for the specific language governing permissions and limitations under
  * the License.
  */
-package zipkin2.stackdriver.translation;
+package zipkin2.translation.stackdriver;
 
-import com.google.common.primitives.UnsignedLongs;
```
adriancole: avoids explicitly declaring a guava version and tracking it

PR #78, opened 2018-05-01
Adds span kinds CONSUMER and PRODUCER

joaoandremartins:
Associates them with SPAN_KIND_UNSPECIFIED on the Stackdriver side,
since, according to CONSUMER and PRODUCER's Javadoc, they don't have
trace IDs, therefore shouldn't be either SpanKind.RPC_CLIENT or
SpanKind.RPC_SERVER.

### DIFF
--- a/translation-stackdriver/src/main/java/zipkin2/translation/stackdriver/SpanTranslator.java
+++ b/translation-stackdriver/src/main/java/zipkin2/translation/stackdriver/SpanTranslator.java
@@ -140,6 +140,8 @@ public final class SpanTranslator {
   }
 
   private static SpanKind getSpanKind(/* Nullable */ Span.Kind zipkinKind) {
+    // Stackdriver Trace still does not have any match for CONSUMER or PRODUCER, and sending it as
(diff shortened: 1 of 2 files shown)


joaoandremartins:
@adriancole FYI

adriancole:
hmm I don't get why UNRECOGNIZED exists if not for things that aren't in stackdriver yet

joaoandremartins:
As-is, the translator is getting `CONSUMER`/`PRODUCER` spans from zipkin and converting them to `UNRECOGNIZED`, which makes Stackdriver Trace error out.

adriancole:
ok maybe add that as the comment for this change? It makes more sense that
we are doing this to avoid a bug vs it be about the trace ID.

a unit test would be nice if you've the time as well.

Thanks!

joaoandremartins:
Comment and unit test added. PTAL

adriancole:
thanks! cutting a release soon

adriancole:
0.4.1 on the way

PR #81, opened 2018-06-05
Implements GCP propagation

joaoandremartins:
This propagation re-uses B3Propagation's injector, while adding a new
CompositeExtractor concept that aims at extracting multiple trace IDs,
in order.

In this case, we're interested in extracting the trace ID from the
'x-cloud-trace-context' header, followed by 'X-B3-TraceId'.

Fixes https://github.com/spring-cloud/spring-cloud-gcp/issues/703

### DIFF
--- a/propagation-stackdriver/pom.xml
+++ b/propagation-stackdriver/pom.xml
@@ -1,0 +1,40 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+
+    Copyright 2016-2018 The OpenZipkin Authors
+
+    Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
+    in compliance with the License. You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+    Unless required by applicable law or agreed to in writing, software distributed under the License
+    is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+    or implied. See the License for the specific language governing permissions and limitations under
+    the License.
+
+-->
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <parent>
+        <artifactId>zipkin-gcp-parent</artifactId>
+        <groupId>io.zipkin.gcp</groupId>
+        <version>0.4.2-SNAPSHOT</version>
+    </parent>
+    <modelVersion>4.0.0</modelVersion>
+
+    <artifactId>zipkin-propagation-stackdriver</artifactId>
+
+    <properties>
+        <main.basedir>${project.basedir}/..</main.basedir>
+    </properties>
+
+    <dependencies>
+        <dependency>
+            <groupId>io.zipkin.brave</groupId>
+            <artifactId>brave</artifactId>
+            <version>5.0.0</version>
+        </dependency>
+    </dependencies>
+</project>
--- a/propagation-stackdriver/pom.xml
+++ b/propagation-stackdriver/pom.xml
@@ -20,11 +20,11 @@
     <parent>
         <artifactId>zipkin-gcp-parent</artifactId>
         <groupId>io.zipkin.gcp</groupId>
-        <version>0.4.2-SNAPSHOT</version>
+        <version>0.5.1-SNAPSHOT</version>
     </parent>
     <modelVersion>4.0.0</modelVersion>
 
-    <artifactId>zipkin-propagation-stackdriver</artifactId>
+    <artifactId>brave-propagation-stackdriver</artifactId>
 
     <properties>
         <main.basedir>${project.basedir}/..</main.basedir>
@@ -34,7 +34,6 @@
         <dependency>
             <groupId>io.zipkin.brave</groupId>
             <artifactId>brave</artifactId>
-            <version>5.0.0</version>
         </dependency>
     </dependencies>
 </project>
--- a/propagation-stackdriver/src/main/java/zipkin2/propagation/stackdriver/CompositeExtractor.java
+++ b/propagation-stackdriver/src/main/java/zipkin2/propagation/stackdriver/CompositeExtractor.java
@@ -37,7 +37,7 @@ public class CompositeExtractor<C> implements TraceContext.Extractor<C> {
     return context;
   }
 
-  public static class FACTORY {
+  public static class Factory {
 
     public static <C> CompositeExtractor<C> newCompositeExtractor(
         TraceContext.Extractor<C>... extractors) {
--- a/propagation-stackdriver/src/main/java/zipkin2/propagation/stackdriver/CompositeExtractor.java
+++ b/propagation-stackdriver/src/main/java/zipkin2/propagation/stackdriver/CompositeExtractor.java
@@ -16,6 +16,10 @@ package zipkin2.propagation.stackdriver;
 import brave.propagation.TraceContext;
 import brave.propagation.TraceContextOrSamplingFlags;
 
+/**
+ * Composite extractor that tries several extractors, in order, to retrieve the tracing context
+ * from a source.
+ */
 public class CompositeExtractor<C> implements TraceContext.Extractor<C> {
 
   private final TraceContext.Extractor<C>[] extractors;
--- a/propagation-stackdriver/src/main/java/zipkin2/propagation/stackdriver/CompositeExtractor.java
+++ b/propagation-stackdriver/src/main/java/zipkin2/propagation/stackdriver/CompositeExtractor.java
@@ -18,22 +18,39 @@ import brave.propagation.TraceContextOrSamplingFlags;
 
 public class CompositeExtractor<C> implements TraceContext.Extractor<C> {
 
-	private TraceContext.Extractor<C>[] extractors;
+  private final TraceContext.Extractor<C>[] extractors;
 
-	public CompositeExtractor(TraceContext.Extractor<C>... extractors) {
-		this.extractors = extractors;
-	}
+  private CompositeExtractor(TraceContext.Extractor<C>... extractors) {
+    this.extractors = extractors;
+  }
 
-	@Override
-	public TraceContextOrSamplingFlags extract(C carrier) {
-		TraceContextOrSamplingFlags context = TraceContextOrSamplingFlags.EMPTY;
+  @Override
+  public TraceContextOrSamplingFlags extract(C carrier) {
+    TraceContextOrSamplingFlags context = TraceContextOrSamplingFlags.EMPTY;
 
-		int currentExtractor = 0;
-		while (context == TraceContextOrSamplingFlags.EMPTY
-				&& currentExtractor < extractors.length) {
-			context = extractors[currentExtractor++].extract(carrier);
-		}
+    int currentExtractor = 0;
+    while (context == TraceContextOrSamplingFlags.EMPTY
+        && currentExtractor < extractors.length) {
+      context = extractors[currentExtractor++].extract(carrier);
+    }
 
-		return context;
-	}
+    return context;
+  }
+
+  public static class FACTORY {
+
+    public static <C> CompositeExtractor<C> newCompositeExtractor(
+        TraceContext.Extractor<C>... extractors) {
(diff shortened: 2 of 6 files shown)


joaoandremartins:
@adriancole can you please take a look at this one? I closed https://github.com/openzipkin/brave/pull/711 as suggested.

I tested this code with a request containing header `x-cloud-trace-context: 8fd836bcfe241ee19a057679a77ba317/4981115762139876185` (similarly to what happens in Google App Engine).
What I observed was that the trace ID is reused, while the span ID of the first request is new.
I think this is expected because in our properties, we enforce `sleuthProperties.setSupportsJoin(false);`.
It would be great if you could comment on this outcome, though.

Besides the lack of unit tests, please let me know what you think should be changed.
I expect you'll have concerns at the performance level (`String.split()`, `String.indexOf()`, etc.). If so, I'm hoping we can iterate on this in master.

FYI @meltsufin @saturnism @artembilan

joaoandremartins:
Also -- any idea why Travis isn't finding `0.4.2-SNAPSHOT` parent in the snapshots repo?

propagation-stackdriver/pom.xml
```diff
@@ -0,0 +1,40 @@
...
+    the License.
+
+-->
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <parent>
+        <artifactId>zipkin-gcp-parent</artifactId>
+        <groupId>io.zipkin.gcp</groupId>
+        <version>0.4.2-SNAPSHOT</version>
```
adriancole: <version>0.5.1-SNAPSHOT</version>

propagation-stackdriver/pom.xml
```diff
@@ -0,0 +1,40 @@
...
+    </parent>
+    <modelVersion>4.0.0</modelVersion>
+
+    <artifactId>zipkin-propagation-stackdriver</artifactId>
```
adriancole: brave-propagation-stackdriver

propagation-stackdriver/pom.xml
```diff
@@ -0,0 +1,40 @@
...
+
+    <properties>
+        <main.basedir>${project.basedir}/..</main.basedir>
+    </properties>
+
+    <dependencies>
+        <dependency>
+            <groupId>io.zipkin.brave</groupId>
+            <artifactId>brave</artifactId>
+            <version>5.0.0</version>
```
adriancole: surprised we don't declare this version

propagation-stackdriver/src/main/java/zipkin2/propagation/stackdriver/CompositeExtractor.java
```diff
@@ -0,0 +1,39 @@
...
+ * the License.
+ */
+package zipkin2.propagation.stackdriver;
+
+import brave.propagation.TraceContext;
+import brave.propagation.TraceContextOrSamplingFlags;
+
+public class CompositeExtractor<C> implements TraceContext.Extractor<C> {
+
+	private TraceContext.Extractor<C>[] extractors;
```
adriancole: final version would hide the ctor final the field and use a factory method instead of ctor. inside that it would guard against null (not guava just throw NPE) or empty (length zero as IAE) and copy the input array.

propagation-stackdriver/src/main/java/zipkin2/propagation/stackdriver/StackdriverTracePropagation.java
```diff
@@ -0,0 +1,146 @@
...
+
+import static brave.internal.HexCodec.lenientLowerHexToUnsignedLong;
+
+public final class StackdriverTracePropagation<K> implements Propagation<K> {
+
+	public static final Propagation.Factory FACTORY = new Propagation.Factory() {
+		@Override public <K> Propagation<K> create(KeyFactory<K> keyFactory) {
+			return new StackdriverTracePropagation<>(keyFactory);
+		}
+
```
adriancole: supports join is most sincerely false in stackdriver ;) we don't allow client and server same id.

propagation-stackdriver/src/main/java/zipkin2/propagation/stackdriver/StackdriverTracePropagation.java
```diff
@@ -0,0 +1,146 @@
...
+	private static final Logger LOG = Logger.getLogger(StackdriverTracePropagation.class.getName());
+
+	private Propagation<K> b3Propagation;
+	private final K traceIdKey;
+	private List<K> fields;
+
+	StackdriverTracePropagation(KeyFactory<K> keyFactory) {
+		this.traceIdKey = keyFactory.create(TRACE_ID_NAME);
+		this.fields = Collections.unmodifiableList(Collections.singletonList(traceIdKey));
+		this.b3Propagation = B3Propagation.FACTORY.create(keyFactory);
```
adriancole: hmm seems this isn't the right place to do this. I would expect the composite to happen above this. for example later stackdriver will likely accept w3c tracecontext format

propagation-stackdriver/src/main/java/zipkin2/propagation/stackdriver/StackdriverTracePropagation.java
```diff
@@ -0,0 +1,146 @@
...
+
+	private static final Logger LOG = Logger.getLogger(StackdriverTracePropagation.class.getName());
+
+	private Propagation<K> b3Propagation;
+	private final K traceIdKey;
+	private List<K> fields;
+
+	StackdriverTracePropagation(KeyFactory<K> keyFactory) {
+		this.traceIdKey = keyFactory.create(TRACE_ID_NAME);
+		this.fields = Collections.unmodifiableList(Collections.singletonList(traceIdKey));
```
adriancole: iirc singletonlist is already unmodifiable. check me?

propagation-stackdriver/src/main/java/zipkin2/propagation/stackdriver/StackdriverTracePropagation.java
```diff
@@ -0,0 +1,146 @@
...
+	}
+
+	@Override public <C> TraceContext.Extractor<C> extractor(Getter<C, K> getter) {
+		if (getter == null) throw new NullPointerException("getter == null");
+		return new CompositeExtractor<>(
+				new XCloudTraceContextExtractor<>(this, getter),
+				b3Propagation.extractor(getter));
+	}
+
+	static final class XCloudTraceContextExtractor<C, K> implements TraceContext.Extractor<C> {
```
adriancole: ok i see. maybe top-level this file and make it public? the "stackdriver" one can mention in docs that it is a composite helper and not to be used inside another composite.

propagation-stackdriver/src/main/java/zipkin2/propagation/stackdriver/StackdriverTracePropagation.java
adriancole: also check traceId128 is true

propagation-stackdriver/src/main/java/zipkin2/propagation/stackdriver/CompositeExtractor.java
```diff
@@ -0,0 +1,39 @@
...
+
+	public CompositeExtractor(TraceContext.Extractor<C>... extractors) {
+		this.extractors = extractors;
+	}
+
```
adriancole: for this to work effectively it needs to be a propagation factory. any member having supportsJoin false makes the whole thing supports join false. similarly on traceid128

propagation-stackdriver/src/main/java/zipkin2/propagation/stackdriver/CompositeExtractor.java
joaoandremartins: Should I be going for something other than arrays, given they don't support generics?

propagation-stackdriver/src/main/java/zipkin2/propagation/stackdriver/CompositeExtractor.java
```diff
@@ -0,0 +1,39 @@
...
+ * Unless required by applicable law or agreed to in writing, software distributed under the License
+ * is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+ * or implied. See the License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package zipkin2.propagation.stackdriver;
+
+import brave.propagation.TraceContext;
+import brave.propagation.TraceContextOrSamplingFlags;
+
```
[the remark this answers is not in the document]
joaoandremartins: Yep, let's move it at the end.

propagation-stackdriver/src/main/java/zipkin2/propagation/stackdriver/CompositeExtractor.java
joaoandremartins: I'm not sure if I understand 100%, but is this accomplished by the `supportsJoin()` returning `false` and `requires128BitTraceId()` returning true in `StackdriverTracePropagation.FACTORY`?

I don't understand how another propagation would fit here, and would be used in conjunction with `StackdriverTracePropagation`, but I might be missing something.

propagation-stackdriver/src/main/java/zipkin2/propagation/stackdriver/StackdriverTracePropagation.java
joaoandremartins: When that happens, wouldn't the right thing be to add new w3c injectors and extractors and add them to the composite returned by `injector()`/`extractor()`?

I guess, conceptually, you could think of `StackdriverTracePropagation` as being _the composite_?

propagation-stackdriver/src/main/java/zipkin2/propagation/stackdriver/CompositeExtractor.java
```diff
@@ -0,0 +1,39 @@
...
+	public CompositeExtractor(TraceContext.Extractor<C>... extractors) {
+		this.extractors = extractors;
+	}
+
+	@Override
+	public TraceContextOrSamplingFlags extract(C carrier) {
+		TraceContextOrSamplingFlags context = TraceContextOrSamplingFlags.EMPTY;
+
+		int currentExtractor = 0;
+		while (context == TraceContextOrSamplingFlags.EMPTY
```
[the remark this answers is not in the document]
joaoandremartins: Hmm interesting, I wasn't aware of `ExtraFieldPropagation`.

1) So, if I'm understanding correctly, if the value of `x-cloud-trace-context` didn't require any special parsing and it was, say, the 32bit hex trace ID, our starter could have just set the `Propagation.Factory` in auto-configuration to `ExtraFieldPropagation.newFactory(B3Propagation.FACTORY, "x-cloud-trace-context")` and we wouldn't need any of this?

2) My reading of your suggestion is that we should document somewhere that once a propagation returns a non-empty context, no more propagations are used. Is that correct? If so, where would be the best place to document this?

In theory, it should be possible to have each propagation modify the context, but potentially overriding settings that the previous propagations set (or conserve them). But do we want to do that?
If we're suggesting that we perform a union operation among the contexts returned by all propagations, how would you suggest that worked?

joaoandremartins:
@adriancole could you PTAL?
Sorry for the pressure. We're sprinting for an M4/RC1 release and this is one of the last couple of bits we'd like to get in for that release.
Thanks!

adriancole:
I am at census workshop today but will look at it between things sure!

On 8 Jun 2018 08:17, "João Martins" <notifications@github.com> wrote:

@adriancole <https://github.com/adriancole> could you PTAL?
Sorry for the pressure. We're sprinting for an M4/RC1 release and this is
one of the last couple of bits we'd like to get in for that release.
Thanks!

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
<https://github.com/openzipkin/zipkin-gcp/pull/81#issuecomment-395793587>,
or mute the thread

propagation-stackdriver/src/main/java/zipkin2/propagation/stackdriver/CompositeExtractor.java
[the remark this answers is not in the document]
adriancole: ok for now, let's make this package private then. can you please add a unit test?

adriancole:
I'll let you do readme post-merge. cutting a release soon (not 1.0.. we'll do that after you verify :))

PR #82, opened 2018-06-11
Adds README for propagation-stackdriver

joaoandremartins:


joaoandremartins:
@adriancole let me know if there is anything missing here!

propagation-stackdriver/README.md
```diff
@@ -0,0 +1,28 @@
+# propagation-stackdriver
+
+This is a propagation that behaves like `B3Propagation`, but also extracts the tracing context from the 'x-cloud-trace-context' key.
```
adriancole: only thing I'd suggest is a code block here showing putting this in your build as from the intro sentence it isn't clear which type people should use and where. so.. make that the next thing and sounds good!

adriancole: sorry I meant to say "builder" ex Tracer.Builder.propagationFactory(what goes here)

adriancole:
great job

PR #83, opened 2018-06-11
Fixes issue where span ID is bigger than long in GAE

joaoandremartins:
App Engine generates span IDs that are longer than a long.
Given that Stackdriver Trace doesn't support trace joins, the incoming
span ID is useless anyway.
So, to save the time of parsing and creating a new long object, we don't
do that and, instead, set 1L as the span ID value, which Brave will
always regenerate anyways.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -36,8 +36,8 @@
     <project.build.resourceEncoding>UTF-8</project.build.resourceEncoding>
 
     <!-- default bytecode version for src/main -->
-    <main.java.version>1.8</main.java.version>
-    <main.signature.artifact>java18</main.signature.artifact>
+    <main.java.version>1.6</main.java.version>
+    <main.signature.artifact>java16</main.signature.artifact>
 
     <main.basedir>${project.basedir}</main.basedir>
 
--- a/pom.xml
+++ b/pom.xml
@@ -36,8 +36,8 @@
     <project.build.resourceEncoding>UTF-8</project.build.resourceEncoding>
 
     <!-- default bytecode version for src/main -->
-    <main.java.version>1.6</main.java.version>
-    <main.signature.artifact>java16</main.signature.artifact>
+    <main.java.version>1.8</main.java.version>
+    <main.signature.artifact>java18</main.signature.artifact>
 
     <main.basedir>${project.basedir}</main.basedir>
 
--- a/propagation-stackdriver/src/main/java/zipkin2/propagation/stackdriver/XCloudTraceContextExtractor.java
+++ b/propagation-stackdriver/src/main/java/zipkin2/propagation/stackdriver/XCloudTraceContextExtractor.java
@@ -54,11 +54,15 @@ public final class XCloudTraceContextExtractor<C, K> implements TraceContext.Ext
       // Try to parse the trace IDs into the context
       TraceContext.Builder context = TraceContext.newBuilder();
       long[] traceId = convertHexTraceIdToLong(tokens[0]);
+      // traceId is null if invalid
       if (traceId != null) {
         boolean traceTrue = true;
 
+        String spanId = "1";
+        // A span ID exists. A TRACE_TRUE flag also possibly exists.
         if (tokens.length >= 2) {
           int semicolonPos = tokens[1].indexOf(";");
+          spanId = semicolonPos == -1 ? tokens[1] : tokens[1].substring(0, semicolonPos);
           traceTrue = semicolonPos == -1
                   || tokens[1].length() == semicolonPos + 4
                   && tokens[1].charAt(semicolonPos + 3) == '1';
@@ -66,7 +70,10 @@ public final class XCloudTraceContextExtractor<C, K> implements TraceContext.Ext
 
         if (traceTrue) {
           result = TraceContextOrSamplingFlags.create(
-                  context.traceIdHigh(traceId[0]).traceId(traceId[1]).spanId(1L).build());
+                  context.traceIdHigh(traceId[0])
+                          .traceId(traceId[1])
+                          .spanId(Long.parseUnsignedLong(spanId))
+                          .build());
         }
       }
     }
--- a/propagation-stackdriver/src/main/java/zipkin2/propagation/stackdriver/XCloudTraceContextExtractor.java
+++ b/propagation-stackdriver/src/main/java/zipkin2/propagation/stackdriver/XCloudTraceContextExtractor.java
@@ -36,8 +36,9 @@ public final class XCloudTraceContextExtractor<C, K> implements TraceContext.Ext
 
   /**
    * Creates a tracing context if the extracted string follows the "x-cloud-trace-context:
-   * TRACE_ID/SPAN_ID" format; or the "x-cloud-trace-context: TRACE_ID/SPAN_ID;0=TRACE_TRUE" format
-   * and {@code TRACE_TRUE}'s value is {@code 1}.
+   * TRACE_ID" or "x-cloud-trace-context: TRACE_ID/SPAN_ID" format; or the
+   * "x-cloud-trace-context: TRACE_ID/SPAN_ID;0=TRACE_TRUE" format and {@code TRACE_TRUE}'s value is
+   * {@code 1}.
    */
   @Override
   public TraceContextOrSamplingFlags extract(C carrier) {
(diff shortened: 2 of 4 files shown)


joaoandremartins:
@meltsufin FYI

propagation-stackdriver/src/main/java/zipkin2/propagation/stackdriver/XCloudTraceContextExtractor.java
```diff
@@ -52,24 +53,20 @@ public TraceContextOrSamplingFlags extract(C carrier) {
 
       // Try to parse the trace IDs into the context
       TraceContext.Builder context = TraceContext.newBuilder();
+      long[] traceId = convertHexTraceIdToLong(tokens[0]);
+      if (traceId != null) {
+        boolean traceTrue = true;
+
+        if (tokens.length >= 2) {
```
adriancole: please put a comment here as to what and why so folks who don't look at git log can easily tell.

propagation-stackdriver/src/main/java/zipkin2/propagation/stackdriver/XCloudTraceContextExtractor.java
```diff
@@ -52,24 +53,20 @@ public TraceContextOrSamplingFlags extract(C carrier) {
...
-          result =
-              TraceContextOrSamplingFlags.create(
-                  context
-                      .traceIdHigh(traceId[0])
-                      .traceId(traceId[1])
-                      .spanId(Long.parseLong(spanId))
-                      .build());
+        if (traceTrue) {
+          result = TraceContextOrSamplingFlags.create(
+                  context.traceIdHigh(traceId[0]).traceId(traceId[1]).spanId(1L).build());
```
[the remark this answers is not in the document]
meltsufin: I'm not sure about the opportunistic parsing because the span ID in Stackdriver is by definition 128-bit. So, overflowing 64 bits is not exactly an edge case.
See: https://cloud.google.com/trace/docs/support

adriancole:
ok well the parent will be invalid which might mess up the UI.

I still think it is worth checking Long.max vs min unless you know all old
stackdriver instrumentation generate 128bit routinely. that way you only
break traces when you have to. I would be surprised if all old stackdriver
code do 128bit span ids but I am sometimes surprised!

Anyway, please test the inpact of this in stackdriver UI.. worth seeing
what would happen as I cant imagine always putting a wrong value will look
better than sometimes or occasionally doing that.

propagation-stackdriver/src/main/java/zipkin2/propagation/stackdriver/XCloudTraceContextExtractor.java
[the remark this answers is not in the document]
meltsufin: @adriancole Sorry, I take back my comment. The span ID is actually 64-bit, but it's unsigned. Hence, it fails about 50% of the time. We just need to deal with it correctly.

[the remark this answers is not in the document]
joaoandremartins:
```
[INFO] Checking unresolved references to org.codehaus.mojo.signature:java16:1.0
[ERROR] /usr/local/google/home/joaomartins/workspaces/zipkin-gcp/propagation-stackdriver/src/main/java/zipkin2/propagation/stackdriver/XCloudTraceContextExtractor.java:75: Undefined reference: long Long.parseUnsignedLong(String)
```

Any reason we're sticking to Java 1.6 here? Or can I change it?

pom.xml
```diff
@@ -36,8 +36,8 @@
     <project.build.resourceEncoding>UTF-8</project.build.resourceEncoding>
 
     <!-- default bytecode version for src/main -->
-    <main.java.version>1.6</main.java.version>
-    <main.signature.artifact>java16</main.signature.artifact>
+    <main.java.version>1.8</main.java.version>
```
adriancole: we still have apps that are JRE 6, so best to revert this.

adriancole:
Here's some code to paste at the bottom of the class. you can use package-private access to add the unit test to the caller's test.

```java

  @Test
  public void parseUnsignedLong() {
    // max int64
    assertThat(parseUnsignedLong("9223372036854775807"))
        .isEqualTo(Long.parseUnsignedLong("9223372036854775807"));

    // max int64 + 1
    assertThat(parseUnsignedLong("9223372036854775808"))
        .isEqualTo(Long.parseUnsignedLong("9223372036854775808"));

    // max uint64
    assertThat(parseUnsignedLong("18446744073709551615"))
        .isEqualTo(Long.parseUnsignedLong("18446744073709551615"));

    // max uint64 + 1
    try {
      parseUnsignedLong("18446744073709551616");
      failBecauseExceptionWasNotThrown(NumberFormatException.class);
    } catch (NumberFormatException e) {
    }
  }

  /** Strictly parses unsigned numbers without a java 8 dependency. */
  static long parseUnsignedLong(String input) throws NumberFormatException {
    if (input == null) throw new NumberFormatException("input == null");
    int len = input.length();
    if (len == 0) throw new NumberFormatException("empty input");
    if (len > 20) throw new NumberFormatException("too long for uint64: " + input);

    // Bear in mind the following:
    // * maximum int64  is  9223372036854775807. Note it is 19 characters
    // * maximum uint64 is 18446744073709551615. Note it is 20 characters

    // It is safe to use defaults to parse <= 18 characters.
    if (len <= 18) return Long.parseLong(input);

    // we now know it is 19 or 20 characters: safely parse the left 18 characters
    long left = Long.parseLong(input.substring(0, 18));

    int digit19 = digitAt(input, 18);
    int rightDigits = 20 - len;
    if (rightDigits == 1) {
      return left * 10 + digit19; // even 19 9's fit safely in a uint64
    }

    int digit20 = digitAt(input, 19);
    int right = digit19 * 10 + digit20;
    // we can run into trouble if the 18 character prefix is greater than the prefix of the
    // maximum uint64, or the remaining two digits will make the number overflow
    // Reminder, largest uint64 is 18446744073709551615
    if (left > 184467440737095516L || (left == 184467440737095516L && right > 15)) {
      throw new NumberFormatException("out of range for uint64: " + input);
    }
    return left * 100 + right; // we are safe!
  }
```

adriancole:
sender-side has java 6, so we should stick with it. I pasted something that
will remove the JRE 8 dependency

joaoandremartins:
thanks @adriancole. PTAL

propagation-stackdriver/src/main/java/zipkin2/propagation/stackdriver/XCloudTraceContextExtractor.java
```diff
@@ -108,4 +108,57 @@ public TraceContextOrSamplingFlags extract(C carrier) {
...
+    if (left > 184467440737095516L || (left == 184467440737095516L && right > 15)) {
+      throw new NumberFormatException("out of range for uint64: " + input);
+    }
+    return left * 100 + right; // we are safe!
+  }
+
+  private static int digitAt(String input, int position) {
+    if (input.length() <= position) throw new NumberFormatException("position out of bounds");
+
+    switch (input.charAt(position)) {
```
adriancole: neat. quicker than subtraction and easy to eyeball

joaoandremartins: Oops I had just replaced it. So I take it you prefer this approach? :P

adriancole: no strong opinion. let's just get this cut!

adriancole:
good work I will cut a release now while still awake

joaoandremartins:
Thanks a mil!

adriancole:
0.6.1 on the way now

joaoandremartins:
Hmm it seems 0.6.1 hasn't been published to Maven Central yet.
Do you know how long it usually takes?

joaoandremartins:
@adriancole sorry for the nagging, but `0.6.1` isn't in Maven Central, yet.
Perhaps you forgot to run a part of the script?
Would it be possible to deploy `0.6.1` to Maven Central today?

We might be able to cut a much needed M4 release on our side today or tomorrow if so. Thanks!

adriancole:
looks like maven central sync failed (not just here but also brave too)

https://bintray.com/openzipkin/maven/zipkin-gcp/0.6.1#central

I will try again but might need to wait until tomorrow as it has been
failing constantly

adriancole:
try 0.6.2 which is aligned with exact same versions of what sleuth GA will have

PR #85, opened 2018-07-07
Migrate to CircleCI 2.0

abesto:
As part of openzipkin/zipkin#2102; based heavily on previous work in openzipkin/brave#725 and openzipkin/zipkin-api#52

Note, unlike other CircleCI builds, this was on Java 8 so far, and remains on Java 8 with this PR.


abesto:
Huh. No CircleCI check was triggered, but on the GH webhooks settings page `6e90d930-81f4-11e8-831a-a7bb0da5951d` seems like CircleCI was notified of the PR. Confused.

abesto:
HAH! Running tests on forked PRs is disabled on https://circleci.com/gh/openzipkin/zipkin-gcp/edit#advanced-settings. Will now enable it, like it is on other projects. The way the build definition and CircleCI secret handling is structured means this is safe, just like for our other repos. Still, shout / disable it if there's some special consideration here I'm not aware of.

.circleci/config.yml
```diff
@@ -0,0 +1,135 @@
...
+
+      - run:
+          name: Tests
+          command: ./mvnw clean verify
+
+      - run:
+          name: Collect test reports
+          command: |
+            mkdir -p /tmp/test-reports/unit-tests/
+            find . -type f -regex ".*/target/surefire-reports/.*xml" -exec cp {} /tmp/test-reports/junit/ \;
```
shakuzen: `/tmp/test-reports/junit/` should be `/tmp/test-reports/unit-tests/`

abesto: Ah. The original was using `junit`, and I managed to mangle things here. Do we keep the naming consistent between `zipkin-gcp` CircleCI 1.0 and CircleCI 2.0, or do we keep the naming consistent between different OpenZipkin projects?

shakuzen: I would personally prefer consistency across projects, but I can understand the argument either way. Particularly in cases where we only have one type of tests, it is less important. I just thought it was nice to distinguish between unit and integration test results, and not focus on the underlying technology, which is the same in the case of JUnit.

abesto:
@shakuzen can haz :eyes: one more time?

PR #91, opened 2018-08-09
Allows ALPN extension when OpenSSL is not available

adriancole:
This allows OpenJDK 1.8 images to live longer (without needing to install OpenSSL). Same approach as Netty uses

See https://github.com/openzipkin/docker-jre-full/issues/9


adriancole:
Thanks @anuraaga for the suggestion!

autoconfigure/storage-stackdriver/src/main/java/zipkin/autoconfigure/storage/stackdriver/ZipkinStackdriverStorageAutoConfiguration.java
```diff
@@ -20,6 +20,7 @@
 import io.grpc.ManagedChannelBuilder;
 import io.grpc.auth.MoreCallCredentials;
 import io.netty.handler.ssl.OpenSsl;
+import io.netty.util.internal.PlatformDependent;
```
adriancole: not concerned on the internal dep as the only consumer of auto-config is the server module (ex this won't taint the deps of those using zipkin-reporter senders)

PR #94, opened 2018-08-29
Shorten tag/label values that are too large for Stackdriver

wjoel:
We had an issue where stackdriver-zipkin (now zipkin-gcp) got stuck in an infinite loop, trying to resend a trace that had a very large SQL statement as a value for one of the tags/labels.

Shorten long tag values to keep within the limits allowed by Stackdriver Trace when extracting labels.


wjoel:
Not that I'm aware of, I can just remove the suffix if you prefer that.

adriancole:
I like the idea of a visual clue of truncation, but otoh I don't
really want to bikeshed on it either. Let's leave it out?

wjoel:
Couldn't agree more, I think it's nice to have but not that important. I've removed it.

adriancole:
thanks tons

PR #97, opened 2018-10-05
Prefer b3 to x-cloud-trace in extractor

### LINKED ISSUE #96: Asymmetry in extractor vs injector breaks propagation
The extractor always prioritizes GCP headers (`X-Cloud-Trace-Context`) over the Brave headers.
However, the injector only propagates traces in the Brave headers. The GCP load-balancer will generate `X-Cloud-Trace-Context` header when one is not present. This causes a break in the trace propagation from service A to B.

I see two possible solutions:
1. Injector should inject both `X-Cloud-Trace-Context` and `X-B3-*` headers.
2. Extractor should prioratize `X-B3-*` headers over `X-Cloud-Trace-Context`.

See: https://github.com/spring-cloud/spring-cloud-gcp/issues/1019

cc/ @saturnism

elefeint:
Addresses mismatch between extractor (which used to prefer X-cloud headers) and injector (which propagates b3 headers) by preferring b3 headers, when available.

Fixes #96 .

### DIFF
--- a/propagation-stackdriver/src/test/java/zipkin2/propagation/stackdriver/StackdriverTracePropagationTest.java
+++ b/propagation-stackdriver/src/test/java/zipkin2/propagation/stackdriver/StackdriverTracePropagationTest.java
@@ -24,49 +24,51 @@ import java.util.Map;
 import static org.assertj.core.api.Assertions.assertThat;
 
 public class StackdriverTracePropagationTest {
-  private static final String XCLOUD_VALUE = "c108dc108dc108dc108dc108dc108d00";
-  private static final String B3_HEADER = "b3";
-  private static final String B3_TRACE_ID = "b3b3b3b3b3b34da6a3ce929d0e0e4736";
-  private static final String B3_VALUE = String.format("%s-00f067aa0ba902b7-1", B3_TRACE_ID);
+  static final String XCLOUD_VALUE = "c108dc108dc108dc108dc108dc108d00";
+  static final String B3_HEADER = "b3";
+  static final String B3_TRACE_ID = "b3b3b3b3b3b34da6a3ce929d0e0e4736";
+  static final String B3_VALUE =  B3_TRACE_ID + "-00f067aa0ba902b7-1";
 
-  private Propagation<String> propagation =
+  Propagation<String> propagation =
       StackdriverTracePropagation.FACTORY.create(Propagation.KeyFactory.STRING);
+  TraceContext.Extractor<Map<String,String>> extractor = propagation.extractor(Map::get);
 
   @Test
   public void b3TakesPrecedenceOverXCloud() {
-    TraceContext.Extractor<String> extractor =
-        propagation.extractor(new FakeGetter(XCLOUD_VALUE, B3_VALUE));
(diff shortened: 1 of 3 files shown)


meltsufin:
Looks good to me. Note that this PR doesn't inject the `x-cloud-trace-context` header -- Item 1 in the issue. It can be done in a separate PR.

meltsufin:
@adriancole Friendly ping.

propagation-stackdriver/src/test/java/zipkin2/propagation/stackdriver/StackdriverTracePropagationTest.java
```diff
@@ -0,0 +1,72 @@
...
+import java.util.HashMap;
+import java.util.Map;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+public class StackdriverTracePropagationTest {
+  private static final String XCLOUD_VALUE = "c108dc108dc108dc108dc108dc108d00";
+  private static final String B3_HEADER = "b3";
+  private static final String B3_TRACE_ID = "b3b3b3b3b3b34da6a3ce929d0e0e4736";
+  private static final String B3_VALUE = String.format("%s-00f067aa0ba902b7-1", B3_TRACE_ID);
```
adriancole: super-nit: normal string cat is probably less distracting here :P

propagation-stackdriver/src/test/java/zipkin2/propagation/stackdriver/StackdriverTracePropagationTest.java
```diff
@@ -0,0 +1,72 @@
...
+
+  private Propagation<String> propagation =
+      StackdriverTracePropagation.FACTORY.create(Propagation.KeyFactory.STRING);
+
+  @Test
+  public void b3TakesPrecedenceOverXCloud() {
+    TraceContext.Extractor<String> extractor =
+        propagation.extractor(new FakeGetter(XCLOUD_VALUE, B3_VALUE));
```
adriancole: I like the idea in general, but I think the FakeGetter makes it less clear, what's going on, vs just using a normal map.

For example, you can initialize an empty mutable map as a field named headers (ps no need to mark fields in a test as private as the keyword is cluttery). You can also initialize a map extractor as a field (ex propagation.extractor(Map::get))

Then, in each test use the actual values you want by setting them in the map. It will look more 

```
headers.put(TRACE_ID_NAME, XCLOUD_VALUE);

TraceContextOrSamplingFlags extracted = extractor.extract(headers);
```

Imho, it reads easier... less distractions, such as "what is the fake getter doing" and "what is this unused object?" It also is much more like normal usage.

elefeint:
@adriancole Thanks for the suggestion; the tests are now much cleaner!

adriancole:
thanks!

elefeint:
@adriancole Could you cut a release for this fix? Then we can pull it into spring-cloud-gcp and close spring-cloud/spring-cloud-gcp#1019.

elefeint:
Thanks!

adriancole:
0.8.3 out and ready cc @anuraaga

PR #102, opened 2018-10-31
Limit trace send waiting to 5 seconds

### LINKED ISSUE #101: Trace sender potentially waits indefinitely for traces to be sent
`AwaitableUnaryClientCallListener` that the trace sender uses to send traces to Stackdriver Trace via GRPC, blocks waiting for `onClose` to be called, potentially indefinitely if there is an error and `onClose` is never called.

Follow-up to #100.

meltsufin:
Fixes #101.

### DIFF
--- a/sender-stackdriver/src/main/java/zipkin2/reporter/stackdriver/internal/AwaitableUnaryClientCallListener.java
+++ b/sender-stackdriver/src/main/java/zipkin2/reporter/stackdriver/internal/AwaitableUnaryClientCallListener.java
@@ -13,6 +13,7 @@
  */
 package zipkin2.reporter.stackdriver.internal;
 
+import com.google.common.annotations.VisibleForTesting;
 import io.grpc.ClientCall;
 import io.grpc.Metadata;
 import io.grpc.Status;
@@ -29,19 +30,21 @@ final class AwaitableUnaryClientCallListener<V> extends ClientCall.Listener<V> {
 
   Object result; // guarded by this
 
-  long timeoutMs = 5000; // how long to wait for server response in milliseconds
+  @VisibleForTesting
+  static long TIMEOUT_MS = 5000; // how long to wait for server response in milliseconds
 
   /**
    * Blocks until {@link #onClose}. Throws if no value was received, multiple
-   * values were received, there was a status error, or waited longer than {@link #timeoutMs}.
+   * values were received, there was a status error, or waited longer than {@link #TIMEOUT_MS}.
    */
   V await() throws IOException {
     boolean interrupted = false;
     try {
       while (true) {
         try {
-          if (!countDown.await(this.timeoutMs, TimeUnit.MILLISECONDS)) {
-            throw new IllegalStateException("Send failed after waiting for " + timeoutMs + "ms.");
+          if (!countDown.await(this.TIMEOUT_MS, TimeUnit.MILLISECONDS)) {
+            throw new IllegalStateException("timeout waiting for onClose. TIMEOUT_MS=" + TIMEOUT_MS
+                + ", resultSet=" + resultSet);
           }
           Object result;
           synchronized (this) {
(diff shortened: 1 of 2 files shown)


meltsufin:
It would be a good idea to add a test for this too, but we before we do that, can you please give feedback on the change? Thanks! @adriancole

sender-stackdriver/src/main/java/zipkin2/reporter/stackdriver/internal/AwaitableUnaryClientCallListener.java
```diff
@@ -28,16 +29,20 @@
 
   Object result; // guarded by this
 
+  long timeoutMs = 5000; // how long to wait for server response in milliseconds
```
adriancole: final unless testing.. if for testing, test it :)

sender-stackdriver/src/main/java/zipkin2/reporter/stackdriver/internal/AwaitableUnaryClientCallListener.java
```diff
@@ -28,16 +29,20 @@
...
+   * values were received, there was a status error, or waited longer than {@link #timeoutMs}.
    */
   V await() throws IOException {
     boolean interrupted = false;
     try {
       while (true) {
         try {
-          countDown.await();
+          if (!countDown.await(this.timeoutMs, TimeUnit.MILLISECONDS)) {
+            throw new IllegalStateException("Send failed after waiting for " + timeoutMs + "ms.");
```
adriancole: "timeout waiting for onClose. timeoutMs=" + timeoutMs + ", valueSet="+valueSet

This will introduce maintenance as maybe there's a weird bug where the value is set, yet we fail anyway (because some bug resulted in onClose never being called). If we find this then there's no point failing I suppose, but we should find out first

sender-stackdriver/src/main/java/zipkin2/reporter/stackdriver/internal/AwaitableUnaryClientCallListener.java
meltsufin: I was thinking of making it configurable, but I just couldn't find a way to easily pass all the way down the chain.

meltsufin:
@adriancole Updated. PTAL.

sender-stackdriver/src/main/java/zipkin2/reporter/stackdriver/internal/AwaitableUnaryClientCallListener.java
```diff
@@ -13,13 +13,15 @@
  */
 package zipkin2.reporter.stackdriver.internal;
 
+import com.google.common.annotations.VisibleForTesting;
```
adriancole: please no guava. a comment will do.

sender-stackdriver/src/main/java/zipkin2/reporter/stackdriver/internal/AwaitableUnaryClientCallListener.java
```diff
@@ -28,16 +30,22 @@
...
+   * values were received, there was a status error, or waited longer than {@link #TIMEOUT_MS}.
    */
   V await() throws IOException {
     boolean interrupted = false;
     try {
       while (true) {
         try {
-          countDown.await();
+          if (!countDown.await(this.TIMEOUT_MS, TimeUnit.MILLISECONDS)) {
+            throw new IllegalStateException("timeout waiting for onClose. TIMEOUT_MS=" + TIMEOUT_MS
```
adriancole: no need to upper underscore the message. if this becomes configurable later we would change the log message without a good reason. upper underscore in the message is distracting imho

adriancole:
thanks a lot

adriancole:
I think you will ask next for release ;)

we are checking about a show stopper on spring boot 2.1 about log4j2. soon as we figure out what to do will release accordingly.

meltsufin:
Thanks for merging. This time the release can actually wait, since this isn't really blocking for us. :)

PR #103, opened 2018-11-07
Handle the TRACE_TRUE option in GCP trace header

dzou:
This adds support to process the TRACE_TRUE option to behave like x-b3-sampled header.

Given a trace of form: `x-cloud-trace-context: TRACE_ID/SPAN_ID;o=TRACE_TRUE`

If TRACE_TRUE
- is `true`: Force the request to be traced
- is `false`: Force the request not to be traced
- is not present: Defer the trace/no-trace decision to the sampler

Fixes https://github.com/spring-cloud/spring-cloud-gcp/issues/749

@meltsufin @elefeint PTAL


dzou:
> I'm going to merge this asking for polish after the fact. @zeagord will be cutting a release, and this looks important to be in that.

Thanks sounds good. I will followup shortly with the polish. In particular I will try to get clarification on the default `span ID = 1` thing to understand where that is coming from.

PR #105, opened 2018-11-12
Cleanup Cloud Trace Context Extractor

dzou:
This cleans up the XCloudTraceContextExtractor following up from #103.

I made a few changes here:
- Instead of setting a default span ID of 1, I set it to 0.

- You can now pass in x-cloud-trace-context in the form of `TRACE_ID/0;o=1`. If 0 is passed in, we wil interpret this as omitting a span ID by returning a `TraceIdContext` object, which reflects the intention described in [Stackdriver docs](https://cloud.google.com/trace/docs/troubleshooting) see section "How do I force a request to be traced?" 

  > SPAN_ID is the decimal representation of the (unsigned) span ID. It should be 0 for the first span in your trace. 

- cleans up the 'o=' options parsing as mentioned in #103.

PTAL @meltsufin

### DIFF
--- a/propagation-stackdriver/src/main/java/zipkin2/propagation/stackdriver/XCloudTraceContextExtractor.java
+++ b/propagation-stackdriver/src/main/java/zipkin2/propagation/stackdriver/XCloudTraceContextExtractor.java
@@ -164,11 +164,11 @@ final class XCloudTraceContextExtractor<C, K> implements TraceContext.Extractor<
     if (optionIndex != -1) {
       String traceTrueValue = traceTrueToken.substring(optionIndex + 2, traceTrueToken.length());
 
-			if (traceTrueValue.equals("1")) {
-				result = true;
-			} else if (traceTrueValue.equals("0")) {
-				result = false;
-			}
+      if (traceTrueValue.equals("1")) {
+        result = true;
+      } else if (traceTrueValue.equals("0")) {
+        result = false;
+      }
     }
 
     return result;
--- a/propagation-stackdriver/src/main/java/zipkin2/propagation/stackdriver/XCloudTraceContextExtractor.java
+++ b/propagation-stackdriver/src/main/java/zipkin2/propagation/stackdriver/XCloudTraceContextExtractor.java
@@ -53,10 +53,11 @@ final class XCloudTraceContextExtractor<C, K> implements TraceContext.Extractor<
       // Try to parse the trace IDs into the context
       TraceContext.Builder context = TraceContext.newBuilder();
       long[] traceId = convertHexTraceIdToLong(tokens[0]);
+
       // traceId is null if invalid
       if (traceId != null) {
         String spanId = "1";
-        Boolean traceTrue = null;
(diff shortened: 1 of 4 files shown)


meltsufin:
LGTM. Does this also solve the problem with the number of spans being different btween `X-B3-` and `X-Cloud-`?

dzou:
> LGTM. Does this also solve the problem with the number of spans being different btween `X-B3-` and `X-Cloud-`?

Unfortunately this does not. I'm going to dig deeper to try to understand the difference.

dzou:
@adriancole, @meltsufin 

If you guys agree with my conclusion on https://github.com/openzipkin/zipkin-gcp/issues/106#issuecomment-438799969 that we are not handling correctly the case where the user sends a spanId == 0 (to signal that they want a root span) I will redo my PR to handle that. That will resolve the problem of the span pointing to a parentSpanId that doesn't exist (which I initially interpreted as a missing span).

meltsufin:
LGTM from me. Nice work getting to the bottom of this missing parent span issue!

propagation-stackdriver/src/main/java/zipkin2/propagation/stackdriver/XCloudTraceContextExtractor.java
```diff
@@ -50,37 +51,43 @@
...
 
-      // Try to parse the trace IDs into the context
-      TraceContext.Builder context = TraceContext.newBuilder();
       long[] traceId = convertHexTraceIdToLong(tokens[0]);
 
       // traceId is null if invalid
       if (traceId != null) {
-        String spanId = "1";
+        long spanId = 0; // 0 indicates no span ID is set by the user
         Boolean traceTrue = null; // null means to defer trace decision to sampler
```
zeagord: Not part of this change but it might be a good idea to replace the `Boolean` to primitive `boolean` with this.

dzou: I thought `Boolean` here was necessary because when this is passed into the traceContext builder it has the following meanings:

- `true`: force request to be traced
- `false`: force request not to be traced
- `null`: Let the sampler decide on whether it should be traced or not.

adriancole: incidentally, the `TraceIdContext` type should be used when you have no span ID. that's probably better than using the incoming trace ID, as it is more to the point.

dzou: @adriancole Right, in this PR I am using the TraceIdContext when I have no span ID. Or am I misunderstanding your comment?

adriancole: you got it.

adriancole:
thanks much!

adriancole:
fyi 0.9.0 includes this

PR #108, opened 2018-11-15
Update README

magikid:
When hitting the health endpoint, the JSON returned includes a details key in the zipkin object.  The details object is what includes the StackdriverStorage object so this updates the argument passed to `jq` so that the status for StackdriverStorage is returned.


adriancole:
thanks much!

PR #112, opened 2018-12-09
Migrate to stackdriver trace API v2.

### LINKED ISSUE #111: Stackdriver Trace V2 API translation
I'm thinking of working on converting to Stackdriver Trace V2 (https://cloud.google.com/trace/docs/reference/v2/rpc/google.devtools.cloudtrace.v2#google.devtools.cloudtrace.v2.TraceService)

One thought I have, is it worth keeping the current V1? I'm not sure if there's a use case for it since GCP itself supports both V1 and V2 let me know if you have any thoughts.

anuraaga:
Took a stab at the migration. Not having to group spans into traces means things become much much simpler. Allocation-wise, since we have to combine trace ID and span ID into a single field for every span now, I think this is the best we can do.

Fixes #111 

Benchmarks. I used a modified benchmark of the parser that serializes the proto since some of the optimizations take advantage of the expectation that the proto will be serialized at some point (using `ByteString` as much as possible, `getSpanIdBytes()`, `UnsafeByteOperations.wrap`). It's still not quite as fast as before, but don't think it would be able to match it considering now a new array must be allocated for every span which concatenates trace ID and span ID and the truncatable strings / attribute values probably add some overhead. As sort of expected, I do notice that the serializing version of the benchmark is ~2x slower on the old code while only ~60% slower with the new one.

New

```
Benchmark                                                                   Mode  Cnt  Score   Error  Units
StackdriverEncoderBenchmarks.encodeClientSpan_json_stackdriver_proto3       avgt   15  1.739 ± 0.037  us/op
StackdriverEncoderBenchmarks.encodeClientSpan_json_zipkin_json              avgt   15  0.755 ± 0.007  us/op
StackdriverEncoderBenchmarks.sizeInBytesClientSpan_json_stackdriver_proto3  avgt   15  0.976 ± 0.027  us/op
StackdriverEncoderBenchmarks.sizeInBytesClientSpan_json_zipkin_json         avgt   15  0.118 ± 0.004  us/op

Benchmark                                   Mode  Cnt    Score   Error  Units
TracesParserBenchmarks.parse100ClientSpans  avgt   15  258.795 ± 3.177  us/op
TracesParserBenchmarks.parseClientSpan      avgt   15    2.579 ± 0.024  us/op
```

Old

```
Benchmark                                                                   Mode  Cnt  Score   Error  Units
StackdriverEncoderBenchmarks.encodeClientSpan_json_stackdriver_proto3       avgt   15  1.796 ± 0.049  us/op
StackdriverEncoderBenchmarks.encodeClientSpan_json_zipkin_json              avgt   15  0.797 ± 0.035  us/op
StackdriverEncoderBenchmarks.sizeInBytesClientSpan_json_stackdriver_proto3  avgt   15  1.162 ± 0.024  us/op
StackdriverEncoderBenchmarks.sizeInBytesClientSpan_json_zipkin_json         avgt   15  0.116 ± 0.001  us/op

Benchmark                                   Mode  Cnt    Score   Error  Units
TracesParserBenchmarks.parse100ClientSpans  avgt   15  157.596 ± 2.184  us/op
TracesParserBenchmarks.parseClientSpan      avgt   15    1.770 ± 0.044  us/op
```

/cc @ewhauser


sender-stackdriver/src/main/java/zipkin2/reporter/stackdriver/StackdriverSender.java
```diff
@@ -164,10 +164,39 @@ public void close() {
     ((ManagedChannel) channel).shutdownNow();
   }
 
-  final class PatchTracesCall extends UnaryClientCall<PatchTracesRequest, Empty> {
+  Span parseTraceIdPrefixedSpan(byte[] traceIdPrefixedSpan) {
```
adriancole: plug.. if anyone gets to the point of groking this or upvoting, we can actually guarantee most span data comes in the same hunk (per trace ID) https://github.com/openzipkin/brave/pull/801

sender-stackdriver/src/main/java/zipkin2/reporter/stackdriver/StackdriverSender.java
```diff
@@ -73,18 +77,26 @@ public StackdriverSender build() {
...
   StackdriverSender(Builder builder) {
     channel = builder.channel;
     callOptions = builder.callOptions;
-    projectId = builder.projectId;
+    projectName = ByteString.copyFromUtf8("projects/" + builder.projectId);
+    traceIdPrefix = projectName.concat(ByteString.copyFromUtf8("/traces/"));
     shutdownChannelOnClose = builder.shutdownChannelOnClose;
-    projectIdFieldSize = 1 /* field no */ + CodedOutputStream.computeStringSizeNoTag(projectId);
+    projectNameFieldSize = 1 /* field no */ + CodedOutputStream.computeBytesSizeNoTag(projectName);
+    spanNameSize = traceIdPrefix.size() + 32 + SPAN_ID_PREFIX.size() + 16;
```
adriancole: the field is named spanNameXX but it seems it is something else. Is it spanIdsSize?

sender-stackdriver/src/main/java/zipkin2/reporter/stackdriver/StackdriverSender.java
```diff
@@ -164,10 +164,39 @@ public void close() {
...
+    byte[] spanName = new byte[spanNameSize];
+    traceIdPrefix.copyTo(spanName, offset);
+    offset += traceIdPrefix.size();
+    System.arraycopy(traceIdPrefixedSpan, 0, spanName, offset, 32);
+    offset += 32;
+    SPAN_ID_PREFIX.copyTo(spanName, offset);
+    offset += SPAN_ID_PREFIX.size();
+    span.getSpanIdBytes().copyTo(spanName, offset);
+
+    span.setNameBytes(UnsafeByteOperations.unsafeWrap(spanName));
```
adriancole: weird so maybe add a comment that span name is actually the trace identifiers :P

translation-stackdriver/src/main/java/zipkin2/translation/stackdriver/SpanTranslator.java
```diff
@@ -49,20 +51,29 @@
...
-    spanBuilder.setKind(kind);
-    spanBuilder.setParentSpanId(parseUnsignedLong(zipkinSpan.parentId()));
-    spanBuilder.setSpanId(parseUnsignedLong(zipkinSpan.id()));
+
+    spanBuilder.setSpanId(zipkinSpan.id());
+    if (zipkinSpan.parentId() != null ) {
+      spanBuilder.setParentSpanId(zipkinSpan.parentId());
+    }
+
+    // NOTE: opencensus prefixes Send. and Recv. based on Kind. For now we reproduce our V1 behavior
```
adriancole: good comment

translation-stackdriver/src/main/java/zipkin2/translation/stackdriver/SpanTranslator.java
```diff
@@ -71,43 +82,22 @@
...
-      return SpanKind.SPAN_KIND_UNSPECIFIED;
-    }
-    if (zipkinKind == Span.Kind.CLIENT) {
-      return SpanKind.RPC_CLIENT;
-    }
-    if (zipkinKind == Span.Kind.SERVER) {
-      return SpanKind.RPC_SERVER;
-    }
-    return SpanKind.UNRECOGNIZED;
+    if (logTranslation) LOG.log(FINE, "<< translated to stackdriver span: {0}", spanBuilder);
```
adriancole: kind should end up somewhere at least making fake annotations.

https://github.com/openzipkin/zipkin/blob/master/zipkin-ui/js/component_ui/spanRow.js#L113

adriancole:
so on cassandra (and elasticsearch) we'll need to ensure the "deduper" is in use to avoid thrashing writes.

In both cases, it might be helpful to reverse-engineer the service-span mapping to re-use the same table. ex `PRIMARY KEY ((type, key), value)` This could make data management in general easier long term.

Since there is a time bomb on elasticsearch #2219, we might want to solve that first before merging this (or at least before cutting a release with it).

Meanwhile, we can allow UI testing to work with static managed list of tags. (ex there may be only several values associated with phase, for example.. so one way is to allow the UI to configure predefined where it is small)

adriancole:
oops last comment was for another issue.. deleting!

anuraaga:
Last mile felt like a hundred miles. But this should finally be ready for a review.

pom.xml
```diff
@@ -43,7 +43,7 @@
 
     <slf4j.version>1.7.25</slf4j.version>
     <!-- only used for protos, we could possibly obviate this if a problem -->
-    <grpc-google-devtools-cloudtrace-v1.version>0.1.1</grpc-google-devtools-cloudtrace-v1.version>
+    <grpc-google-cloud-trace-v2.version>0.38.0</grpc-google-cloud-trace-v2.version>
```
adriancole: does this pull any weird deps? just check you don't need to make any exclusions

ewhauser:
This is good to be merged from my perspective.

I tracked down the issue with the tags. It's a bug in the Zipkin JS library which is sending tags as binaryAnnotations to the v2 endpoint. I'll open an issue over there.

adriancole:
@ewhauser thanks and "eek" about the tag bug!

PR #113, opened 2018-12-17
avoid conflicts with upstream dependencies by using shaded netty; thi…

ewhauser:
…s will make openzipkin/zipkin#2331 easier


adriancole:
just need to check that boringssl thing still works, which I assume would just yeah need to verify by using against stackdriver

adriancole:
testing locally now

ewhauser:
Should have mentioned I'm still testing it. Wiring it up now

ewhauser:
First Angular traces written to Stackdriver... 

screen shot 2018-12-16 at 6 31 00 pm

adriancole:
I also verified here with spring cloud sleuth. LGTM

Issue #116, opened 2019-02-20
Update 0.9.x to use grpc-google-cloud-trace-v1


saturnism:
Zipkin GCP 0.10.x is using grpc-google-cloud-trace-v2 Stackdriver Trace v2 API is not yet GA.
Zipkin GCP 0.9.0 is using devtools which is deprecated and pulls in old dependencies. It would be great to update a 0.9.x version that continues to use v1 API but w/ newer dependencies.

adriancole:
I think my higher concern is maintaining a release branch.. we dont do that
generally.. is it possible for these to coexist (package names different)?

saturnism:
package names are in fact different. it's possible to have both `StackdriverSender`s

adriancole:
I suspect it would be a V1 encoder if we did it this way..
https://github.com/openzipkin/zipkin-gcp/blob/master/sender-stackdriver/src/main/java/zipkin2/reporter/stackdriver/StackdriverEncoder.java

saturnism:
chatted w/ @meltsufin; we think we can put this on hold for now. will re-open if situation changes.

PR #117, opened 2019-02-20
moved trace proto dependency to grpc-google-loud-trace-v1

saturnism:
for #116 but may need to rebase onto a 0.9.x branch.


adriancole:
should we close this?

adriancole:
made an assumption this is more or less dead :P

PR #120, opened 2019-03-23
Use armeria-grpc client in storage-stackdriver.

### LINKED ISSUE #119: armerify
I think we can do more and get out of dodging netty and/or boringssl conflicts by moving the stackdriver collector (not necessarily the sender) to armeria. This would imply figuring out if armeria has an oauth feature which is compatible with stackdriver

cc @openzipkin/armeria

anuraaga:
One concern is the API change in storage-stackdriver's `Builder` not sure what the API guarantee on that is.

Only ran integration tests.

Fixes #119

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -179,20 +179,22 @@
         <artifactId>grpc-google-cloud-trace-v2</artifactId>
         <version>${grpc-google-cloud-trace-v2.version}</version>
         <exclusions>
+          <!-- let collector (armeria) or sender (grpc) control the protobuf versions -->
           <exclusion>
             <groupId>io.grpc</groupId>
             <artifactId>grpc-protobuf</artifactId>
           </exclusion>
           <exclusion>
-            <groupId>io.grpc</groupId>
-            <artifactId>grpc-stub</artifactId>
+            <groupId>com.google.protobuf</groupId>
+            <artifactId>protobuf-java</artifactId>
           </exclusion>
           <exclusion>
-            <groupId>com.google.guava</groupId>
-            <artifactId>guava</artifactId>
+            <groupId>com.google.api</groupId>
+            <artifactId>api-common</artifactId>
           </exclusion>
         </exclusions>
       </dependency>
+
       <dependency>
         <groupId>com.google.guava</groupId>
         <artifactId>guava</artifactId>
--- a/pom.xml
+++ b/pom.xml
@@ -47,15 +47,14 @@
 
     <!-- must match (ex. use spring-boot used by zipkin-server) -->
     <spring-boot.version>2.1.3.RELEASE</spring-boot.version>
+    <armeria.version>0.83.0</armeria.version>
     <zipkin.version>2.12.7</zipkin.version>
     <zipkin-reporter.version>2.8.2</zipkin-reporter.version>
     <brave.version>5.6.3</brave.version>
 
-    <!-- must match https://github.com/grpc/grpc-java/blob/master/SECURITY.md#netty -->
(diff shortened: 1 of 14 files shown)


anuraaga:
Hmm - does the build require updating the license headers to 2019?

adriancole:
we dont have api guarantees pre 1.0!

adriancole:
almost sorted

adriancole:
ok this works now, except we have to bump zipkin's guava as grpc no longer works with guava 19. this requires a release of zipkin before the integration will work via curl downloading etc.

storage-stackdriver/src/main/java/zipkin2/storage/stackdriver/StackdriverSpanConsumer.java
```diff
@@ -82,17 +81,21 @@ public BatchWriteSpansCall clone() {
...
+              @Override public void onSuccess(Empty empty) {
+                callback.onSuccess(empty);
+              }
+
+              @Override public void onFailure(Throwable throwable) {
+                callback.onError(throwable);
+              }
+            },
+            MoreExecutors.directExecutor());
+      } catch (RuntimeException | Error e) {
```
anuraaga: I guess this is equivalent to `Throwable`

sender-stackdriver/pom.xml
```diff
@@ -42,17 +42,6 @@
     <dependency>
       <groupId>com.google.api.grpc</groupId>
       <artifactId>grpc-google-cloud-trace-v2</artifactId>
-      <version>${grpc-google-cloud-trace-v2.version}</version>
-      <exclusions>
```
anuraaga: Was there something wrong with these exclusions? gRPC should always have compatibility even with stubs compiled against an older version - if they aren't that's quite surprising...

pom.xml
```diff
@@ -170,6 +175,31 @@
...
+          <exclusion>
+            <groupId>io.grpc</groupId>
+            <artifactId>grpc-protobuf</artifactId>
+          </exclusion>
+          <exclusion>
+            <groupId>io.grpc</groupId>
+            <artifactId>grpc-stub</artifactId>
+          </exclusion>
+          <exclusion>
+            <groupId>com.google.guava</groupId>
```
anuraaga: Filed https://github.com/googleapis/gapic-generator/issues/2648 ...

anuraaga: Though related to it, I think it should be ok to remove `api-common` instead of this deeper one.

sender-stackdriver/pom.xml
adriancole: it was more about the version skew. we can test if removing the exclusions doesnt whack the tree

storage-stackdriver/src/main/java/zipkin2/storage/stackdriver/StackdriverSpanConsumer.java
adriancole: oops meant to throw this e heh. in such case throwable would jank the signature

sender-stackdriver/pom.xml
adriancole: oh these are in the parent pom now.. that's what happened.

pom.xml
adriancole: revised. thanks!

adriancole:
ok I tested against live stackdriver and all good

adriancole:
thanks for the hard work @anuraaga!

adriancole:
out at 0.11.0 (into the docker, too)

trustin:
Nice work, @anuraaga !

PR #122, opened 2019-04-02
Add config for server timeout

sergeymeshkov:
PR for #122


sender-stackdriver/src/main/java/zipkin2/reporter/stackdriver/internal/UnaryClientCall.java
```diff
@@ -23,16 +23,28 @@
 import zipkin2.Callback;
 
 public abstract class UnaryClientCall<ReqT, RespT> extends Call.Base<RespT> {
+  public static final int DEFAULT_SERVER_TIMEOUT_MS = 5000;
   final ClientCall<ReqT, RespT> call;
   final ReqT request;
+  final long serverTimeoutMs;
 
   protected UnaryClientCall(
```
adriancole: this is an internal constructor, so better to break it than maintain two imho

sender-stackdriver/src/test/java/zipkin2/reporter/stackdriver/internal/UnaryClientCallTimeoutTest.java
```diff
@@ -0,0 +1,46 @@
...
+  private static final long SERVER_TIMEOUT_MS = 50;
+
+  @Before
+  public void setUp() {
+    server.getServiceRegistry().addService(traceService);
+    call = new BatchWriteSpansCall(server.getChannel(), BatchWriteSpansRequest.newBuilder().build(), SERVER_TIMEOUT_MS);
+  }
+
+  @Test(expected = IllegalStateException.class)
+  public void execute_timeout() throws Throwable {
```
adriancole: rather than splitting the test into 3 files, which can be distracting to this change.. I would recommend just overwriting the call instance like so..

```
int overriddenTimeout = 100;
call = new BatchWriteSpansCall(server.getChannel(), BatchWriteSpansRequest.newBuilder().build(), overriddenTimeout);
...
                 Thread.sleep(overriddenTimeout + 5);
```

This keeps the focus of the change where intended and removes a need for a flurry of formatting requests.

sender-stackdriver/src/test/java/zipkin2/reporter/stackdriver/internal/UnaryClientCallTest.java
```diff
@@ -46,32 +47,37 @@
...
 
   BatchWriteSpansCall call;
 
   @Before
-  public void setUp() throws Throwable {
+  public void setUp() {
     server.getServiceRegistry().addService(traceService);
-    call = new BatchWriteSpansCall(server.getChannel(), BatchWriteSpansRequest.newBuilder().build());
+    call = new BatchWriteSpansCall(server.getChannel(), BatchWriteSpansRequest.newBuilder().build(), DEFAULT_SERVER_TIMEOUT_MS);
+  }
```
adriancole: there's fuzz in this file. for example, this turns into a redundant thing below. also there are bunch of accidental indents...

```
  @Before
  public void setUp() {
    server.getServiceRegistry().addService(traceService);
    call = new BatchWriteSpansCall(server.getChannel(), BatchWriteSpansRequest.newBuilder().build(), DEFAULT_SERVER_TIMEOUT_MS);
  }

  public void setUp(long serverResponseTimeout) {
    server.getServiceRegistry().addService(traceService);
    call = new BatchWriteSpansCall(server.getChannel(), BatchWriteSpansRequest.newBuilder().build(), serverResponseTimeout);
}
```

sergeymeshkov: Are you saying that these two "identical" method calls should be removed? At the moment, the default timeout is 5 seconds, if you remove this configuration, the test execution will take 5 seconds + timeout. Is it works?

adriancole: ah sorry you are right, I didn't notice the subtle difference. I would dispatch the setup method or just inline `setUp(long)` as it causes a mental glitch (in me.. but maybe only me?)

adriancole:
ps thanks for your patience with me. if ever you feel like.. man.. this is a bore! let me know as I can de-fuzz on the way in.

sergeymeshkov:
> ps thanks for your patience with me. if ever you feel like.. man.. this is a bore! let me know as I can de-fuzz on the way in.

No problem :) 
Thank you for your recommendations. I think it’s our common task to do everything as best we can

adriancole:
thanks!

PR #124, opened 2019-05-15
Use UnaryGrpcClient in storage-stackdriver instead of gRPC stubs.

anuraaga:
Now `storage-stackdriver` doesn't depend on gRPC or guava.

I was a bit unsure what to do with it's protobuf dependency as armeria-grpc-protocol doesn't depend on it so it wouldn't control the version. I've never seen an incompatible API in protobuf before though and recommend just pinning to latest version like here.

### DIFF
--- a/autoconfigure/storage-stackdriver/src/main/java/zipkin/autoconfigure/storage/stackdriver/ZipkinStackdriverStorageAutoConfiguration.java
+++ b/autoconfigure/storage-stackdriver/src/main/java/zipkin/autoconfigure/storage/stackdriver/ZipkinStackdriverStorageAutoConfiguration.java
@@ -15,15 +15,14 @@ package zipkin.autoconfigure.storage.stackdriver;
 
 import com.google.auth.Credentials;
 import com.google.auth.oauth2.GoogleCredentials;
-import com.google.common.io.ByteStreams;
 import com.linecorp.armeria.client.ClientFactory;
 import com.linecorp.armeria.client.ClientOptionsBuilder;
+import com.linecorp.armeria.client.HttpClient;
+import com.linecorp.armeria.common.HttpMethod;
+import com.linecorp.armeria.common.RequestHeaders;
 import io.netty.handler.ssl.OpenSsl;
 import io.netty.util.internal.PlatformDependent;
 import java.io.IOException;
-import java.io.InputStream;
-import java.net.HttpURLConnection;
-import java.net.URL;
 import java.util.Collections;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.beans.factory.annotation.Qualifier;
@@ -35,8 +34,6 @@ import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.Configuration;
 import zipkin2.storage.StorageComponent;
 import zipkin2.storage.stackdriver.StackdriverStorage;
-
-import static com.google.common.base.Preconditions.checkState;
 
 @Configuration
 @EnableConfigurationProperties(ZipkinStackdriverStorageProperties.class)
@@ -61,19 +58,19 @@ public class ZipkinStackdriverStorageAutoConfiguration {
     }
     try {
       return getDefaultProjectId();
-    } catch (IOException exception) {
+    } catch (Throwable t) {
       throw new IllegalArgumentException("Missing required property: projectId");
     }
   }
 
-  String getDefaultProjectId() throws IOException {
-    URL url = new URL("http://metadata.google.internal/computeMetadata/v1/project/project-id");
-    HttpURLConnection connection = (HttpURLConnection) url.openConnection();
-    connection.setRequestProperty("Metadata-Flavor", "Google");
-    connection.setRequestMethod("GET");
-    try (InputStream responseStream = connection.getInputStream()) {
-      return new String(ByteStreams.toByteArray(responseStream), "UTF-8");
-    }
+  String getDefaultProjectId() {
+    HttpClient client = HttpClient.of("http://metadata.google.internal/");
+    return client.execute(RequestHeaders.of(
+        HttpMethod.GET, "/computeMetadata/v1/project/project-id",
+        "Metadata-Flavor", "Google"))
+        .aggregate()
+        .join()
+        .contentUtf8();
   }
 
   @Bean
@@ -90,8 +87,11 @@ public class ZipkinStackdriverStorageAutoConfiguration {
       ClientFactory clientFactory,
       ZipkinStackdriverStorageProperties properties,
       Credentials credentials) {
-    checkState(OpenSsl.isAvailable() || jettyAlpnAvailable(),
-        "OpenSsl or ALPN is required. This usually requires either JDK9+, jetty-alpn, or netty-tcnative-boringssl-static");
+    if (!OpenSsl.isAvailable() && !jettyAlpnAvailable()) {
+      throw new IllegalStateException(
+          "OpenSsl or ALPN is required. This usually requires either JDK9+, jetty-alpn, or "
+              + "netty-tcnative-boringssl-static");
+    }
     return StackdriverStorage.newBuilder(properties.getApiHost())
         .projectId(projectId)
         .strictTraceId(strictTraceId)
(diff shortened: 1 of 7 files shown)


autoconfigure/storage-stackdriver/src/main/java/zipkin/autoconfigure/storage/stackdriver/ZipkinStackdriverStorageAutoConfiguration.java
```diff
@@ -61,19 +58,19 @@ String projectId() {
     }
     try {
       return getDefaultProjectId();
-    } catch (IOException exception) {
+    } catch (Throwable t) {
```
adriancole: will need an import to keep this compiling :P

```suggestion
    } catch (Throwable t) {
      Call.propagateIfFatal(t);
```

autoconfigure/storage-stackdriver/src/main/java/zipkin/autoconfigure/storage/stackdriver/ZipkinStackdriverStorageAutoConfiguration.java
```diff
@@ -61,19 +58,19 @@ String projectId() {
...
-  String getDefaultProjectId() throws IOException {
-    URL url = new URL("http://metadata.google.internal/computeMetadata/v1/project/project-id");
-    HttpURLConnection connection = (HttpURLConnection) url.openConnection();
-    connection.setRequestProperty("Metadata-Flavor", "Google");
-    connection.setRequestMethod("GET");
-    try (InputStream responseStream = connection.getInputStream()) {
-      return new String(ByteStreams.toByteArray(responseStream), "UTF-8");
-    }
+  String getDefaultProjectId() {
+    HttpClient client = HttpClient.of("http://metadata.google.internal/");
```
adriancole: cooooooool

storage-stackdriver/pom.xml
```diff
@@ -37,22 +37,37 @@
...
 
     <dependency>
       <groupId>com.google.api.grpc</groupId>
-      <artifactId>grpc-google-cloud-trace-v2</artifactId>
+      <artifactId>proto-google-cloud-trace-v2</artifactId>
+    </dependency>
+
+    <dependency>
+      <groupId>com.google.protobuf</groupId>
+      <artifactId>protobuf-java</artifactId>
```
adriancole: /me thinking.. no need to change anything

suppose we need this as long as we aren't using something else to generate the protos for cloud trace (ex the library above). While not required, it would be neat to use wire or something to generate these as they are simple types and the dependency tree is less frantic. That said, I think we'd be in the same boat eventually due to oauth etc, so probably not important.

pom.xml
```diff
@@ -45,12 +45,13 @@
 
     <!-- matching armeria/grpc/zipkin -->
     <spring-boot.version>2.1.4.RELEASE</spring-boot.version>
-    <armeria.version>0.83.0</armeria.version>
+    <armeria.version>0.85.0</armeria.version>
```
adriancole: note this is later than last released zipkin. we'll probably not want to release this until they are synced (likely 0.86)

anuraaga:
In that case, do you think we should hold off on merging until zipkin's released?

adriancole:
one thing we can do to release tension is release master now against the last released version of zipkin.

I do think that merging this now, out of sync, could be confusing to whoever (if any) are testing off master...

anuraaga:
Yeah - will hold off on merging then until things get synced up

adriancole:
I pushed this with latest zipkin (armeria etc) when integrating it, I noticed this when debug is up

```
2019-06-14 23:09:50.993 DEBUG 78023 --- [-worker-nio-2-6] z.s.i.ZipkinHttpCollector                : Cannot store spans [94d3046ff03e0726/7c83c828057eb693] due to CompletionException(com.linecorp.armeria.common.grpc.protocol.ArmeriaStatusException: Non-successful HTTP response code: 404 Not Found)
```

adriancole:
```bash
curl -sSL https://zipkin.io/quickstart.sh | bash -s
./mvnw clean install
mv ./autoconfigure/storage-stackdriver/target/zipkin-autoconfigure-storage-stackdriver-0.12.1-SNAPSHOT-module.jar stackdriver.jar
GOOGLE_APPLICATION_CREDENTIALS=$HOME/.gcp/credentials.json STORAGE_TYPE=stackdriver STACKDRIVER_PROJECT_ID=zipkin-demo     java     -Dloader.path='stackdriver.jar,stackdriver.jar!/lib'     -Dspring.profiles.active=stackdriver     -cp zipkin.jar     org.springframework.boot.loader.PropertiesLauncher  --logging.level.zipkin2=DEBUG
```

then in another window hit brave-webmvc-example or similar. it is probably something small

adriancole:
tidied up the tree under this change. still getting the 404.. looking into it carefully now

anuraaga:
Ah I remembered that some Google services don't respect the gRPC wire protocol... Can't remember if trace was one but probably was.

Probably need to add a decorator to replace the content type header like this one.

https://github.com/curioswitch/curiostack/blob/master/common/google-cloud/core/src/main/java/org/curioswitch/curiostack/gcloud/core/grpc/GrpcApiClientBuilder.java#L59

adriancole:
ok the above is a successful request prior to this change. Notably it has `params` and the scheme is set to `gproto+h2`, where as the latter is a request from this change, lacking `params` in the log output (maybe ok?) and with a generic scheme of `none+h2`. Maybe there's something else to notice which I've not.

Anyway, @anuraaga hope you can sort this given the breadcrumbs :P

Before:
```
Request: {startTime=2019-06-15T03:09:46.790Z(1560568186790975), length=401B, duration=9282µs(9282089ns), scheme=gproto+h2, headers=[:method=POST, :path=/google.devtools.cloudtrace.v2.TraceService/BatchWriteSpans, :authority=cloudtrace.googleapis.com, :scheme=https, content-type=application/grpc, grpc-accept-encoding=gzip, grpc-timeout=15000000u, authorization=Bearer XXXXXXXXXXXXXX, user-agent=armeria/0.87.0], content=DefaultRpcRequest{serviceType=GrpcLogUtil, method=google.devtools.cloudtrace.v2.TraceService/BatchWriteSpans, params=[name: "projects/zipkin-demo"
spans {
  name: "projects/zipkin-demo/traces/0000000000000000d6cf8fb3eff42182/spans/5609cf662b573bf3"
  span_id: "5609cf662b573bf3"
  parent_span_id: "d6cf8fb3eff42182"
  display_name {
    value: "get /api"
  }
  start_time {
    seconds: 1560568185
    nanos: 145467000
  }
  end_time {
    seconds: 1560568185
    nanos: 152601000
  }
  attributes {
    attribute_map {
      key: "/kind"
      value {
        string_value {
          value: "server"
        }
      }
    }
    attribute_map {
      key: "/http/method"
      value {
        string_value {
          value: "GET"
        }
      }
    }
    attribute_map {
      key: "http.path"
      value {
        string_value {
          value: "/api"
        }
      }
    }
    attribute_map {
      key: "mvc.controller.class"
      value {
        string_value {
          value: "Backend"
        }
      }
    }
    attribute_map {
      key: "mvc.controller.method"
      value {
        string_value {
          value: "printDate"
        }
      }
    }
    attribute_map {
      key: "endpoint.ipv4"
      value {
        string_value {
          value: "192.168.0.104"
        }
      }
    }
    attribute_map {
      key: "/component"
      value {
        string_value {
          value: "backend"
        }
      }
    }
  }
}
]}}
```

Now:
```
Request: {startTime=2019-06-15T03:06:11.965Z(1560567971965318), length=403B, duration=8863µs(8863589ns), scheme=none+h2, headers=[:method=POST, :path=/google.devtools.cloudtrace.v2.TraceService/BatchWriteSpans, :authority=cloudtrace.googleapis.com, :scheme=https, content-type=application/grpc+proto, content-length=403, authorization=Bearer XXXXXXXXXXXXXX, user-agent=armeria/0.87.0]}
```

adriancole:
oic yeah before it was "application/grpc" whereas in this change it is "application/grpc+proto". I can try to change it

adriancole:
that was it, thanks!

adriancole:
works now

adriancole:
fyi if we want to go further in a later change (not that we must) the rest of the dodgy deps are around google's oauth library. With that gone it is all clean. Anyway this is a big difference.. thanks!

adriancole:
going out now as 0.13.0

PR #126, opened 2019-05-16
Allow running zipkin-gcp on Java 9+ without ALPN

anuraaga:
Happened to notice this when wanting to copy this code into armeria


adriancole:
thanks!

PR #128, opened 2019-06-15
Updates to latest zipkin

adriancole:


adriancole:
tested local and is ok

PR #136, opened 2019-08-17
Check Stackdriver sender health

### LINKED ISSUE #131: Better check() method for StackdriverSender
Currently, the `StackdriverSender.check` method is just a placeholder.

```java
  public CheckResult check() {
    return CheckResult.OK;
  }
```

Follow-up to https://github.com/spring-cloud/spring-cloud-gcp/issues/1752.

elefeint:
Check whether Stackdriver is functioning by sending a malformed request and checking the response for a known validation error, configurable through the builder. Successful response from Stackdriver, while unexpected, is considered successful.

Background:
It turns out that I lied about a malformed request validation being in front of the service check -- I had been using the client library to test before, which is where the pre-validations came from. So using a malformed request that never reaches the user's stackdriver console works great. I also have a version stashed that sends a full and proper trace, but that no longer seems necessary. [EDIT: curiously, not sending any spans, just a project ID, results in a successful gRPC call even if API is disabled, so I am sticking with sending one empty span].

Obsolete comment due to decision to not cache status:

> Because `sendSpans` returns and does *not* execute the call, I could not use its outcome to set the healthcheck (since the execution happens a level up, in reporter, and reporter sending check result back to Sender seems backwards). I looked into chaining a `handleError()` call a bit, but it looks like only one error handler is expected, best to leave it to the end user. If nothing calls it, `setCheckResult()` is unnecessary. 

Fixes #131.


sender-stackdriver/src/main/java/zipkin2/reporter/stackdriver/ExpiringHealthCheck.java
```diff
@@ -0,0 +1,102 @@
...
+import java.time.Clock;
+import java.util.function.Function;
+import zipkin2.Call;
+import zipkin2.CheckResult;
+
+/**
+ * Stackdriver Trace Health check that throttles server calls to once per minute.
+ * Capable of accepting recent status to avoid making unnecessary calls.
+ */
+public class ExpiringHealthCheck {
```
adriancole: since this is internal, we can make this package-private.

sender-stackdriver/src/main/java/zipkin2/reporter/stackdriver/ExpiringHealthCheck.java
```diff
@@ -0,0 +1,102 @@
...
+
+  private Clock clock = Clock.systemDefaultZone();
+
+  private final ByteString projectName;
+
+  private final Function<BatchWriteSpansRequest, Call> serviceCall;
+
+  private CheckResult lastCheckResult;
+
+  private long expirationTimestamp;
```
adriancole: the thread could be different, so probably volatile is good here.

sender-stackdriver/src/main/java/zipkin2/reporter/stackdriver/ExpiringHealthCheck.java
```diff
@@ -0,0 +1,102 @@
...
+    this.serviceCall = serviceCall;
+  }
+
+  /**
+   * Tests Stackdriver Trace by sending a malformed request.
+   * @return successful check result if input validation error is detected; failed check with the
+   * reported exception otherwise.
+   */
+  public CheckResult check() {
+    if (clock.millis() > expirationTimestamp) {
```
adriancole: System.nanoTime will be more reliable. There are edge cases about it because the value returned can be negative https://github.com/openzipkin/brave/blob/5a17fc018613958db00cc7b6951826e95f1b9d6c/brave/src/test/java/brave/sampler/RateLimitingSamplerTest.java#L48

adriancole:
derp.. of course this is sender not consumer :P sorry .. but anyway some sort of wrapping approach, whichever api we are actually wrapping, will likely get the job done.

anuraaga:
Just wondering, do we need the expiration? I think if a user configures a particular health check period they'd expect each check to be fresh, and they'd have to keep the two intervals (this expiration and e.g., a kubernetes health check) in sync somehow. I think the expiration was probably important when we thought of using a separate, possibly low-quota IAM API but if just sending a cheap empty trace, any thoughts of not having an expiration?

adriancole:
ps ooop sorry.. handling the error is a lot simpler by decorating the call like this.

```java
   return new BatchWriteSpansCall(request.build())
       .handleError((throwable, callback) -> {
         lastThrowable = throwable; // remember the failure
         callback.onError(throwable);
       })
       .map(empty -> {
         lastThrowable = null; // success
         return null;
       });
```

adriancole:
@anuraaga not everyone uses the health check periodically. so far we aren't talking about a cheap empty trace or any billing event, rather tracking the last error. I don't think the logic or code is complex enough to warrant punting to side effects on the account. my 2p

adriancole:
@anuraaga you might have missed that stackdriver has some internal guards that skip empty traces here #131. We do use the "empty trace" trick elsewhere, but so far I'm told that won't work here.

anuraaga:
Ah I was having trouble understanding the decorator stuff but finally realized you're suggesting a totally different approach of not making extra requests :)

adriancole:
it is fair enough to question if we want to treat this like a gauge instead, hold the last value indefinitely. I suppose if expiring proves to be complex it would still be better to hold an error too long than not show an error at all. I didn't really expect expiration to be something debatable, but it is a fair question.

adriancole:
@anuraaga yep this one is a bit weird, though it isn't that bad an options. some of our checks aren't that good. it got me thinking if we shouldn't share errors in other places or not. this could be a decent experiment towards that as we don't have another signal

anuraaga:
@adriancole Yeah saw that - from this PR description though it sounds like that was mistaken, hence the current implementation is using a dummy request in `sendTestSpan` from what I can tell.

adriancole:
sorry about my distractions, @elefeint and @anuraaga  I should have read the descriptions

yep, we should just make a malformed call, possibly save it off and .clone().execute() it on the health check like we do others. This focus on the only thing of importance which is catching the correct error :)

adriancole:
in other words, remembering the exception is a distraction unless there's a billing effect of this approach.

adriancole:
I'll delete my invalid code advice notes here.

sender-stackdriver/src/main/java/zipkin2/reporter/stackdriver/ExpiringHealthCheck.java
```diff
@@ -0,0 +1,102 @@
...
+
+    Span testSpan = Span.newBuilder().build();
+
+    BatchWriteSpansRequest request = BatchWriteSpansRequest.newBuilder()
+      .setNameBytes(projectName)
+      .addSpans(testSpan)
+      .build();
+
+    try {
+      serviceCall.apply(request).execute();
```
adriancole: nit: you could save this call of as a field and use testCall.clone().execute();

since we expect an error you might still want to invert the logic. ex use an error handler into success, yet map success into error as passing would be unexpected?

adriancole:
Here are some responses to the description, which I should have read instead of jumping right into the code earlier and causing distractions.

> This is what I have so far; needs to be Java6-ified, and I am sure my style settings were off. To misquote Scarlett O'Hara, I will think about it on Monday.
yep only thing really is the Clock type, but I think we can possibly delete expiration at this point as no-where else do we rate-limit the health check.

> It turns out that I lied about a malformed request validation being in front of the service check -- I had been using the client library to test before, which is where the pre-validations came from. So using a malformed request that never reaches the user's stackdriver console works great. I also have a version stashed that sends a full and proper trace, but that no longer seems necessary.
This was the part I should have read, I am embarrassed and apologize for wasting comments before. This is great about malformed request. I just made some notes that because what is malformed and the behavior around it could drift.. possibly we need to test success doesn't accidentally occur. That or add a comment that success is ok because we don't care if stackdriver eats malformed, and don't want a late crash later if we are too strict. either way, since we'd be testing an odd practice, a one-liner on the success and error block seems about right. 


> Because sendSpans returns and does not execute the call, I could not use its outcome to set the healthcheck (since the execution happens a level up, in reporter, and reporter sending check result back to Sender seems backwards). I looked into chaining a handleError() call a bit, but it looks like only one error handler is expected, best to leave it to the end user. If nothing calls it, setCheckResult() is unnecessary.
I wouldn't design this for the end user to know how this is implemented at least from a code level. This allows us to change it later if a health check api happens. Instead, I'd rely on this being the only call site. When calling .execute(), really no big advantage to using handleError as the callback only helps in async. To answer your question though, handleError is wrapping, so you could do this an unlimited amount of times until you run out of stack :P

elefeint:
"possibly we need to test success doesn't accidentally occur" -- that seems okay to me. If Stackdriver removes the validation, that's on them to catch, but as far as Zipkin sender is concerned, the underlying service is alive and well.

The validation error message drifting does worry me, since error messages can be fairly casually changed (I just did a drive-by change to a pubsub message last week). I am going to generalize from checking for specific message to catching the INVALID_ARGUMENT prefix, which should catch any "this is a validation error, therefore you at minimum got past coarse API failures into actual functionality".

And @anuraaga , thank you for saving me from overengineering  :)

adriancole:
ps I'm about to sign off, but if you add that as a property to the sender builder, then someone in zipkin-gcp or otherwise could change the pattern in a pinch. reduces re-release pressure.

until next time! thanks @elefeint

adriancole:
I mean spring-cloud-gcp.. demasiado gcps :P

sender-stackdriver/src/main/java/zipkin2/reporter/stackdriver/StackdriverSender.java
```diff
@@ -55,6 +55,7 @@ public static Builder newBuilder(Channel channel) { // visible for testing
     CallOptions callOptions = DEFAULT;
     boolean shutdownChannelOnClose;
     long serverResponseTimeoutMs = DEFAULT_SERVER_TIMEOUT_MS;
+    String expectedHealthcheckError = "INVALID_ARGUMENT";
```
anuraaga: Is it possible to cast the exception to StatusException or such to extract the actual gRPC status code?

sender-stackdriver/src/main/java/zipkin2/reporter/stackdriver/StackdriverSender.java
```diff
@@ -161,8 +178,26 @@ public int messageSizeInBytes(int traceIdPrefixedSpanSize) {
...
+      healthcheckCall.clone().execute();
+    } catch (Exception e) {
+      if (e.getMessage().contains(expectedHealthcheckError)) {
+        return CheckResult.OK;
+      }
+      return CheckResult.failed(e);
+    }
+
+    // Stackdriver should not execute the malformed call successfully, so the code is unlikely to
+    // get here. But it happens, that means the server is responding, and so is minimally healthy.
```
anuraaga: I think we can be more explicit for future maintainers, something like `Currently the rpc throws an exception if there is a problem, which we handle above. If we get here despite a problem, the implementation changed and we need to update this check. It's unlikely enough that we can wait and see.`.

sender-stackdriver/src/main/java/zipkin2/reporter/stackdriver/StackdriverSender.java
```diff
@@ -32,9 +32,9 @@
 import zipkin2.reporter.Sender;
 import zipkin2.reporter.stackdriver.internal.UnaryClientCall;
 
-import static zipkin2.reporter.stackdriver.internal.UnaryClientCall.DEFAULT_SERVER_TIMEOUT_MS;
 import static com.google.protobuf.CodedOutputStream.computeUInt32SizeNoTag;
 import static io.grpc.CallOptions.DEFAULT;
+import static zipkin2.reporter.stackdriver.internal.UnaryClientCall.DEFAULT_SERVER_TIMEOUT_MS;
```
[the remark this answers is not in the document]
elefeint: Autoformat; I'll revert since it's not related tho this PR.

sender-stackdriver/src/main/java/zipkin2/reporter/stackdriver/StackdriverSender.java
elefeint: Yes; much better than comparing text.

sender-stackdriver/src/main/java/zipkin2/reporter/stackdriver/StackdriverSender.java
elefeint: I'll update the wording, but should there be some logging/metrics for this? Otherwise nobody will ever notice the behavior change.

elefeint:
I've tested with a service account lacking permissions, and `check()` behaves correctly (propagates gRPC exception with `Status.PERMISSION_DENIED`).

@adriancole @anuraaga do you think this PR is good to go?

sender-stackdriver/src/main/java/zipkin2/reporter/stackdriver/StackdriverSender.java
```diff
@@ -161,8 +171,28 @@ public int messageSizeInBytes(int traceIdPrefixedSpanSize) {
...
+   */
   @Override
   public CheckResult check() {
+    try {
+      healthcheckCall.clone().execute();
+    } catch (StatusRuntimeException sre) {
+      if (sre.getStatus().getCode() == Status.Code.INVALID_ARGUMENT) {
+        return CheckResult.OK;
+      }
+      return CheckResult.failed(sre);
```
adriancole: not sure the SRE team will like being called a failure!

Issue #130, opened 2019-07-30
Add automated testing for zipkin-gcp


saturnism:
Add automated testing for zipkin-gcp against a stackdriver project.

- [x] Stackdriver Sender #140 #141
- [x] Stackdriver Reporter #143 
- [x] Against Zipkin Server #150 
- [x] Docker image #159

adriancole:
Then you can emulate the existing zipkin-example-foo tests by sending a generated trace https://github.com/openzipkin/zipkin-js-example/blob/master/.circleci/config.yml

then you can use the rest api of stackdriver to check, or just check /metrics instead if that's too hard

adriancole:
cc @openzipkin/devops-tooling

meltsufin:
Are we using Circle CI or Travis for testing this project? I only see Travis config and it skips test.
I can add a service account key to either for enabling some integration tests against Stackdriver.
Also, which tests would you want to run?

adriancole:
we still check on pull requests. agree that if there were service related
things that would drift, or sneaky commits that were never tested on a
branch, it could be a problem. The skipping test on master could be
disabled on this repo, as the tests aren't flakey (heh yet!)

https://github.com/openzipkin/zipkin-gcp/blob/master/travis/publish.sh#L116-L118

We'd need a new smoke test unless whatever we configure for that job is
available (eg keys). Then someone can always run "manually" but it is still
less manual than really manual (eg me doing things manually literally by
assembling a server).

I would suggest composing the server together instead of running the things
within a java classpath as often this is where things break. ex like the
docker image https://github.com/openzipkin/docker-zipkin-gcp/. In fact we
could just do this at the docker repo.. basically use jitpack to get the
latest server and the lastest stackdriver module and fail if we can't send
a trace and then look it back up with gcp api.

While I referred to a circleci job, I wouldn't suggest employing circleci
as we already use travis anyway (I'd prefer thing mentioned to not use
circleci, just not bothered enough to port on own right now.).

For this sort of test probably a periodic could be helpful, for lack of
smarter interval/signal and also not wanting a chance to lock-up accounts.
https://docs.travis-ci.com/user/cron-jobs/

If we end up testing things "once removed" ex via a docker job in another
repo, we just mention at the README where that is.. I think it is ok,
especially as many use docker and failing there is what I'm usually nervous
about.

sg?

meltsufin:
[...]

PR #137, opened 2019-08-23
Enable existing JUnit 4 tests and run integration tests on verify

saturnism:
For #130

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -233,6 +233,12 @@
       <version>4.12</version>
     </dependency>
     <dependency>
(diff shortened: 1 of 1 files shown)


saturnism:
Confirmed that the unit tests ran: https://travis-ci.org/openzipkin/zipkin-gcp/builds/575959699#L2424

Also, because `publish.sh` runs `verify`, so the existing integration test also ran: https://travis-ci.org/openzipkin/zipkin-gcp/builds/575959699#L2440

adriancole:
Thanks @saturnism!

PR #140, opened 2019-08-26
Stackdriver integration tests

saturnism:
#130 

- Renamed existing `ITStackdriverSender`test to `AsyncReporterStackdriverSenderTest.java`
- Updated `ITStackdriverSender` to test against a service account
- Configure the service account path via `GOOGLE_APPLICATION_CREDENTIALS` env var


sender-stackdriver/src/test/java/zipkin2/reporter/stackdriver/AsyncReporterStackdriverSenderTest.java
```diff
@@ -0,0 +1,102 @@
...
+
+import static java.util.Arrays.asList;
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.mockito.Matchers.any;
+import static org.mockito.Mockito.doAnswer;
+import static org.mockito.Mockito.never;
+import static org.mockito.Mockito.spy;
+import static org.mockito.Mockito.verify;
+
+/** Same as ITStackdriverSpanConsumer: tests everything wired together */
```
adriancole: rename the other one? I agree this is not an IT compared to running against real. Ex in normal Zipkin we use IT as a way to indicate slow tests that can be flakey. gRPC tests are middle-ground as they do open sockets and can be flakey but also aren't as slow or deserving skips as running real.

adriancole:
I have a change in brave I've neglected for a long time so look at this closer after that. thanks for the help!

sender-stackdriver/src/test/java/zipkin2/reporter/stackdriver/ITStackdriverSender.java
```diff
@@ -13,90 +13,131 @@
...
-                  return null;
-                })
-        .when(traceService)
-        .batchWriteSpans(any(BatchWriteSpansRequest.class), any(StreamObserver.class));
+  @Test
+  public void healthcheckFailNoPermission() {
+    CheckResult result = reporterNoPermission.check();
+    assertThat(result.ok()).isFalse();
+    assertThat(result.error()).isNotNull();
+    assertThat(result.error()).isInstanceOf(StatusRuntimeException.class);
```
anuraaga: Can combine this and next line into `isInstanceOfSatisfying` if you want.

sender-stackdriver/src/test/java/zipkin2/reporter/stackdriver/ITStackdriverSender.java
```diff
@@ -13,90 +13,131 @@
...
   AsyncReporter<Span> reporter;
+  AsyncReporter<Span> reporterNoPermission;
+  TraceServiceGrpc.TraceServiceBlockingStub traceServiceGrpcV1;
+  Channel v1Channel;
 
   @Before
-  public void setUp() {
-    server.getServiceRegistry().addService(traceService);
+  public void setUp() throws IOException {
+  	// Application Default credential is configured using the GOOGLE_APPLICATION_CREDENTIALS env var
```
anuraaga: Maybe ` or gcloud SDK when run locally.`

.travis.yml
```diff
@@ -1,48 +1,33 @@
...
-      - https://webhooks.gitter.im/e/ead3c37d57527214e9f2
-      - https://webhooks.gitter.im/e/e57478303f87ecd7bffc
+    - https://webhooks.gitter.im/e/ead3c37d57527214e9f2
+    - https://webhooks.gitter.im/e/e57478303f87ecd7bffc
     on_success: change
     on_failure: always
-
 env:
   global:
-    # Ex. travis encrypt BINTRAY_USER=your_github_account
```
anuraaga: Looks like we lost these comments which are usefulI I think. The new secret could have a comment about what the name of the service account is.

sender-stackdriver/src/test/java/zipkin2/reporter/stackdriver/ITStackdriverSender.java
adriancole: I'd prefer to see an `assumeThat` here which will ignore the test vs break the build, when remote stuff is not around. That way you can craft a better error. PS I think the build is broke not sure intended or not.

saturnism: TIL :D updated to use `assumeThatCode` since `GoogleCredentials.getApplicationDefault` throws an exception that we need to catch.

adriancole:
something still seems busted..

saturnism:
it looks like the assumption was determined not met

```
org.junit.AssumptionViolatedException: assumption was not met due to: 
Expecting code not to raise a throwable but caught
```

but it's still a failure rather than a skip.

adriancole:
if me, I'd check the state it needs instead as the error message isn't 100% clear anyway. there's a slight drift risk, but then if there was, we'd have to reconfig the job

saturnism:
even if i tried a simple `assumeThat(System.getenv("GOOGLE_APPLICATION_CREDENTIALS")).isNotBlank();`, in intelliJ it shows yellow X, but is marked as a test failure, and in maven, it's marked as failure rather than skip.

saturnism:
ok i think i found the issue...

adriancole:
looks like you did. good detective work again @saturnism. Thanks for all the work.

Thanks for the review @elefeint and @anuraaga!

PR #143, opened 2019-09-03
Added integration test for Stackdriver Storage with autoconfiguration

saturnism:
For #130 

Added integration test for Zipkin Stackdriver Reporter, which is a different path than the Stackdriver Sender.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,7 +18,8 @@ before_install:
   # Encrypted service account file for integration testing
   # This file is referenced via GOOGLE_APPLICATION_CREDENTIALS env var
   # Ex. travis encrypt-file travis/zipkin-gcp-ci-0d7917f58da7.json
-  - openssl aes-256-cbc -K $encrypted_0d0e64b78c84_key -iv $encrypted_0d0e64b78c84_iv -in travis/zipkin-gcp-ci-0d7917f58da7.json.enc -out travis/zipkin-gcp-ci-0d7917f58da7.json -d
+  # Decryption will fail on pull requests. Allow failures to build the PRs.
+  - openssl aes-256-cbc -K $encrypted_0d0e64b78c84_key -iv $encrypted_0d0e64b78c84_iv -in travis/zipkin-gcp-ci-0d7917f58da7.json.enc -out travis/zipkin-gcp-ci-0d7917f58da7.json -d || true
 
 # Override default travis to use the maven wrapper; skip license on travis due to #1512
 install: ./mvnw install -DskipTests=true -Dlicense.skip=true -Dmaven.javadoc.skip=true -B -V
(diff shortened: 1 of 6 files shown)


saturnism:
Good thing I'm doing this w/ a PR outside of `zipkin-gcp` repo ;) caught quite a few issues w/ travis and PR. Fixed those in this commit too.

pom.xml
```diff
@@ -49,9 +49,10 @@
     <zipkin.version>2.16.2</zipkin.version>
     <zipkin-reporter.version>2.10.2</zipkin-reporter.version>
     <brave.version>5.6.10</brave.version>
-    <grpc.version>1.22.1</grpc.version>
+    <grpc.version>1.22.2</grpc.version>
```
saturnism: this is updated to match armeria's grpc version.

.travis.yml
```diff
@@ -18,7 +18,8 @@ before_install:
   # Encrypted service account file for integration testing
   # This file is referenced via GOOGLE_APPLICATION_CREDENTIALS env var
   # Ex. travis encrypt-file travis/zipkin-gcp-ci-0d7917f58da7.json
-  - openssl aes-256-cbc -K $encrypted_0d0e64b78c84_key -iv $encrypted_0d0e64b78c84_iv -in travis/zipkin-gcp-ci-0d7917f58da7.json.enc -out travis/zipkin-gcp-ci-0d7917f58da7.json -d
```
elefeint: In spring-cloud-gcp, we [explicitly gate](https://github.com/spring-cloud/spring-cloud-gcp/blob/0513b8784f8828918aacacf18db80a992ae4134c/.travis.yml#L24) on $TRAVIS_SECURE_ENV_VARS. In principle, TRAVIS_PULL_REQUEST could also be used to conditionally decrypt as it's false when not running on a pull request.

Environment variable reference: https://docs.travis-ci.com/user/environment-variables/

sender-stackdriver/pom.xml
```diff
@@ -98,7 +98,7 @@
     <dependency>
       <groupId>org.awaitility</groupId>
       <artifactId>awaitility</artifactId>
-      <version>3.1.6</version>
+      <version>${awaitability.version}</version>
```
elefeint: s/awaitability/awaitility

saturnism: good catch!

.travis.yml
saturnism: great idea; i'll do the same.

saturnism: there is a similar gate for decrypting file: https://github.com/spring-cloud/spring-cloud-gcp/blob/0513b8784f8828918aacacf18db80a992ae4134c/.travis.yml#L37

i've updated here to use the same gate logic.

Issue #123, opened 2019-04-23
Google Stackdriver Trace drops spans without name


TYsewyn:
When enabling debug logging for the `zipkin2` package when using Spring Cloud Sleuth (Greenwich.SR1) we get following error:
```
Dropped 18 spans due to StatusRuntimeException(INVALID_ARGUMENT: TraceSpan.name for TraceSpan.span_id 1829621750604935860 cannot be empty.)
io.grpc.StatusRuntimeException: INVALID_ARGUMENT: TraceSpan.name for TraceSpan.span_id 1829621750604935860 cannot be empty.
at io.grpc.Status.asRuntimeException(Status.java:526)
at zipkin2.reporter.stackdriver.internal.AwaitableUnaryClientCallListener.onClose(AwaitableUnaryClientCallListener.java:96)
at io.grpc.PartialForwardingClientCallListener.onClose(PartialForwardingClientCallListener.java:39)
at io.grpc.ForwardingClientCallListener.onClose(ForwardingClientCallListener.java:23)
at io.grpc.ForwardingClientCallListener$SimpleForwardingClientCallListener.onClose(ForwardingClientCallListener.java:40)
at io.grpc.internal.CensusStatsModule$StatsClientInterceptor$1$1.onClose(CensusStatsModule.java:678)
at io.grpc.PartialForwardingClientCallListener.onClose(PartialForwardingClientCallListener.java:39)
at io.grpc.ForwardingClientCallListener.onClose(ForwardingClientCallListener.java:23)
at io.grpc.ForwardingClientCallListener$SimpleForwardingClientCallListener.onClose(ForwardingClientCallListener.java:40)
at io.grpc.internal.CensusTracingModule$TracingClientInterceptor$1$1.onClose(CensusTracingModule.java:397)
at io.grpc.internal.ClientCallImpl.closeObserver(ClientCallImpl.java:459)
at io.grpc.internal.ClientCallImpl.access$300(ClientCallImpl.java:63)
at io.grpc.internal.ClientCallImpl$ClientStreamListenerImpl.close(ClientCallImpl.java:546)
at io.grpc.internal.ClientCallImpl$ClientStreamListenerImpl.access$600(ClientCallImpl.java:467)
at io.grpc.internal.ClientCallImpl$ClientStreamListenerImpl$1StreamClosed.runInContext(ClientCallImpl.java:584)
at io.grpc.internal.ContextRunnable.run(ContextRunnable.java:37)
[...]

PR #144, opened 2019-09-03
Set Stackdriver Span displayName to unknown if Zipkin Span name is empty

saturnism:
#123


saturnism:
The tests will run properly once #143 is merged.

Issue #148, opened 2019-09-13
Span without service name causes NullPointerException


z-oo:
Example:
```
curl -X POST -s localhost:9411/api/v2/spans -H'Content-Type: application/json' -d '[{"traceId":"fe1583ed5be1b96cf40538f958ca97a6","id":"96169d26c53f25ab","name":"loader::unload","timestamp":1568360755828060,"duration":124,"localEndpoint":{"ipv6":"::1"}}]'
```
result in
```
java.lang.NullPointerException: null
	at zipkin2.translation.stackdriver.AttributesExtractor.extract(AttributesExtractor.java:83) ~[zipkin-translation-stackdriver-0.14.1-SNAPSHOT.jar!/:?]
	at zipkin2.translation.stackdriver.SpanTranslator.translate(SpanTranslator.java:112) ~[zipkin-translation-stackdriver-0.14.1-SNAPSHOT.jar!/:?]
	at zipkin2.translation.stackdriver.SpanTranslator.translate(SpanTranslator.java:60) ~[zipkin-translation-stackdriver-0.14.1-SNAPSHOT.jar!/:?]
	at zipkin2.storage.stackdriver.StackdriverSpanConsumer.accept(StackdriverSpanConsumer.java:51) ~[zipkin-storage-stackdriver-0.14.1-SNAPSHOT.jar!/:?]
	at zipkin2.collector.Collector.store(Collector.java:200) ~[zipkin-collector-2.16.3-SNAPSHOT.jar!/:?]
	at zipkin2.collector.Collector$StoreSpans.run(Collector.java:229) ~[zipkin-collector-2.16.3-SNAPSHOT.jar!/:?]
	at com.linecorp.armeria.common.AbstractRequestContext.lambda$makeContextAware$1(AbstractRequestContext.java:69) ~[armeria-0.91.0.jar!/:?]
	at com.linecorp.armeria.common.AbstractRequestContext.lambda$makeContextAware$1(AbstractRequestContext.java:69) ~[armeria-0.91.0.jar!/:?]
	at io.micrometer.core.instrument.AbstractTimer.record(AbstractTimer.java:160) ~[micrometer-core-1.2.1.jar!/:1.2.1]
	at io.micrometer.core.instrument.Timer.lambda$wrap$0(Timer.java:144) ~[micrometer-core-1.2.1.jar!/:1.2.1]
[...]

PR #149, opened 2019-09-13
Make serviceName check null-safe

z-oo:
Span object stores `localEndpoint.serviceName` empty string as null so `.isEmpty()` throws NullPointerException.


z-oo:
This should fix #148

z-oo:
@devinsba test added. PTAL.

PR #150, opened 2019-09-14
Test with Zipkin Server

saturnism:


saturnism:
PTAL at the approach for zipkin server test.

The actual call to start this test will be moved to run iff on mater branch

travis/publish.sh
```diff
@@ -104,6 +104,50 @@ safe_checkout_master() {
   fi
 }
 
+test_server() {
+  # Test Stackdriver Storage module with the latest version of Zipkin server.
+  temp_dir=$(mktemp -d)
+  pushd $temp_dir
+
+  # Download and unpack Zipkin Server
+  wget https://jitpack.io/com/github/openzipkin/zipkin/zipkin-server/master-SNAPSHOT/zipkin-server-master-SNAPSHOT-exec.jar
```
anuraaga: Let's use the sonatype snapshots repo instead.

travis/publish.sh
```diff
@@ -104,6 +104,50 @@ safe_checkout_master() {
...
+
+  # Zipkin server JAR is a self-executable with headers that causes unzip to exit with status 1
+  # Fix the archive first.
+  zip -A zipkin-server-master-SNAPSHOT-exec.jar
+  unzip zipkin-server-master-SNAPSHOT-exec.jar -d zipkin
```
anuraaga: I think a similar server test could be done by using maven project and adding the snapshot repo and dependency on the exec jar. Then you could just call ZipkinServer.main. A lot of the logic here like the massive wait-for-it would be simpler in Java.

What do you think of that approach?

travis/publish.sh
```diff
@@ -104,6 +104,50 @@ safe_checkout_master() {
...
+
+  # Unpack our module
+  unzip -n $TRAVIS_BUILD_DIR/autoconfigure/storage-stackdriver/target/*-module.jar -d zipkin
```
anuraaga: I think you can set both jars to `-cp` instead of unzipping

travis/publish.sh
saturnism: i'm having some trouble finding where this is :( a pointer would be appreciate it!

travis/publish.sh
saturnism: that didn't work for me. I was wondering why unzip was necessary in Zipkin's Dockerfiles, but apparently, can't do it w/ simple `java -cp`

travis/publish.sh
saturnism: moving this back to the issue for more discussion. https://github.com/openzipkin/zipkin-gcp/issues/130#issuecomment-531521645

adriancole: I don't like the dependency on zipkin server approach. as it sends the wrong signal about supported integrations and also doesn't test the actual integration as we support it.

adriancole: for example those on gitter know we are plagued with ad-hoc questions about custom servers via dependencies, to the point of us wanting to refactor our entire README strategy. Please don't test via dependencies as A. people will get the wrong idea  B. it hides integration problems C. it doesn't follow our README.

travis/publish.sh
adriancole: unzip is used in zipkin's dockerfiles as it speeds up bootstrapping and allows layering to play games with the contents.

travis/publish.sh
adriancole: I don't agree with using sonatype as it implies maintenance to get the latest snapshot version. Let's only do that if we must.

adriancole: for example, it looks like this, which is unnecessary blinding

```
    # the current version of zipkin-lens is in a path of the same name in a jar file. This extracts it.
    LENS_BASE=$ZIPKIN_REPO/io/zipkin/zipkin-lens && \
    LENS_VERSION=$(curl -sSL $LENS_BASE/maven-metadata.xml | sed -n '/<version>/s/.*<version>\([^<]*\)<\/version>.*/\1/p'|tail -1) && \
    LENS_BUILD=$(curl -sSL $LENS_BASE/$LENS_VERSION/maven-metadata.xml | sed -n '/<value>/s/.*<value>\([^<]*\)<\/value>.*/\1/p'| head -1) && \
    curl -sSL $LENS_BASE/$LENS_VERSION/zipkin-lens-$LENS_BUILD.jar > zipkin-lens.jar && \
```

https://github.com/openzipkin/docker-zipkin/blob/master/zipkin-ui/Dockerfile#L10-L14

If we end up with problems in jitpack, we can revisit, but please let's keep as much cruft out as possible.

travis/publish.sh
adriancole: I dont know where my comment is, but I wholly do not like the idea of call ZipkinServer.main. I anti-like it, actively dislike it as it is doesn't actually test integration as in our README. We should test what we support

travis/publish.sh
adriancole: I'm surprised what didn't work, but we can try to pin down what it was. I think in one scenario we had to unzip due to a nested compressed jar problem? @jeqo do you remember?

https://github.com/openzipkin-contrib/zipkin-storage-kafka/pull/30

travis/publish.sh
adriancole: apologies about the redundant comment.. github refresh wasn't refreshing. I'll delete the last one

travis/publish.sh
anuraaga: We've already had problems with running the wrong version of a jitpack build without specifying a hash. The caching behavior is too opaque to put in an automated test IMO.

travis/publish.sh
anuraaga: The main advantage would be if we want to run more tests than just the health, which would be much simpler I'm Java. But if just a health check gives us confidence then I think this is fine too.

travis/publish.sh
anuraaga: Ah yeah the spring loader.path property doesn't support nested jars which is annoying. But that's mainly a problem for the zipkin-server docker image which supports arbitrarily named jars as plugins. Since we know the names here I thought `-cp` should work without issues (I don't unzip when doing this sort of command locally) but not a big deal to unzip anyways.

travis/publish.sh
adriancole: I'm not positive that was jitpack's fault. you and I were flipping between many things and with not much rest. It is entirely possible the flipping with no sleep was more an issue than jitpack iotw

travis/publish.sh
adriancole: I understood the goal of this to be a smoke test, not a replacement for the IT stuff we do already. Ideally, we don't have to cover too many paths in this style.

travis/publish.sh
adriancole: We have nested jars in other repos. -cp isn't the integration path for this, it is what is in the readme (loader path) which is capable of nested jars. IIRC there was a problem with "compressed" jars, but that wasn't entirely validated either.

I routinely run the README for example, which had been working, so either we're not noticing something subtle here or there's an env thing..

travis/publish.sh
adriancole: ps either way translates to the same.. if someone was too tired and this didn't work or similar because they didn't force docker to invalidate cache etc.. I'm fine removing jitpack from this step if you feel strongly about it. we will still have the docker cache miss thing on any snapshot though.. even what I pasted scraping xml.

tldr I'm ok either way on this part

travis/publish.sh
```diff
@@ -104,6 +104,50 @@ safe_checkout_master() {
...
+  ZIPKIN_PID=$!
+
+  echo "Waiting for Zipkin server to start..."
+  $TRAVIS_BUILD_DIR/travis/wait-for-it.sh localhost:9411 -t 60
+  exit_status=$?
+  if [ $exit_status -ne 0 ]; then
+    exit $exit_status
+  fi
+
+  health_check_result=$(curl --silent localhost:9411/actuator/health| jq -r .details.zipkin.status)
```
elefeint: nit: add space after |

travis/publish.sh
```diff
@@ -104,6 +104,50 @@ safe_checkout_master() {
...
+
+  if [ "$health_check_result" != "UP" ]; then
+    echo "Health check failed!"
+    curl --silent localhost:9411/actuator/health | jq .
```
elefeint: Does the error message work as expected? The unquoted period in `jq .` could be interpreted by shell as a current directory. Maybe surround it with single quotes to guarantee no interpretation: ` jq '.'`.

travis/publish.sh
saturnism: done!

travis/publish.sh
saturnism: i think this works as expected. `jq -h` shows a similar usage:
```
$ echo '{"foo": 0}' | jq .
...
```

Once I fix the wait-for-it stuff, I'll do a test that should fail the health check and we can validate the error msg.

saturnism:
> ... Apparently it's packaged for both Debian and Ubuntu, so maybe `apt-get install`-ing it would be cleaner?

Good call. Will update.

saturnism:
> > ... Apparently it's packaged for both Debian and Ubuntu, so maybe `apt-get install`-ing it would be cleaner?
> 
> Good call. Will update.

oops - not in ubuntu xenial :(

saturnism:
good news - i think i got the script to run the integration test pretty well.
bad news - it's failing :D

https://travis-ci.org/openzipkin/zipkin-gcp/builds/585941140#L4988

```
Exception in thread "armeria-common-blocking-tasks-1-1" java.lang.NoClassDefFoundError: io/grpc/Context
	at io.opencensus.trace.CurrentSpanUtils.getCurrentSpan(CurrentSpanUtils.java:37)
	at io.opencensus.trace.Tracer.spanBuilder(Tracer.java:308)
	at com.google.api.client.http.HttpRequest.execute(HttpRequest.java:847)
	at com.google.auth.oauth2.ServiceAccountCredentials.refreshAccessToken(ServiceAccountCredentials.java:358)
	at com.google.auth.oauth2.OAuth2Credentials.refresh(OAuth2Credentials.java:165)
	at com.google.auth.oauth2.OAuth2Credentials.getRequestMetadata(OAuth2Credentials.java:151)
	at com.google.auth.Credentials.blockingGetToCallback(Credentials.java:113)
	at com.google.auth.Credentials$1.run(Credentials.java:99)
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1128)
	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:628)
	at java.base/java.lang.Thread.run(Thread.java:834)

```

travis/publish.sh
saturnism: i updated the command line to use the jar. it was my bad i didn't fully understand how to refer to all the jars differently. ptal the new command line.

adriancole:
@saturnism I raised a PR to fix the classpath thing

travis/publish.sh
adriancole: your negative test worked. I removed it!

adriancole:
ok pass.. now I'm going to un-vendor wait-for-it using a stable copy we control. https://raw.githubusercontent.com/openzipkin-contrib/wait-for-it/master/wait-for-it.sh

travis/publish.sh
adriancole: tabled as the shell scripting to do the alternative is gross #155

travis/publish.sh
adriancole: most of our problems have been related to things we now do in health check, so let's try this and re-evalutate if it is not pulling weight

travis/publish.sh
adriancole: changed to match the README

adriancole:
Cool stuff @saturnism. Just verified master, and I think we can finally cut a release safely :)

Thanks for all the support on this @anuraaga @elefeint @abesto 

```
+curl --silent localhost:9411/info

+jq .

2019-09-26 13:55:44.316  INFO 5527 --- [           main] z.s.ZipkinServer                         : Started ZipkinServer in 5.535 seconds (JVM running for 7.22)

2019-09-26 13:55:44.508  INFO 5527 --- [orker-epoll-2-2] c.l.a.i.JavaVersionSpecific              : Using version specific APIs from Java9VersionSpecific

{

  "zipkin": {

    "version": "2.17.1-SNAPSHOT",

    "commit": "844c43b"

  }

}

++curl --silent localhost:9411/health

++jq -r .status

+health_check_result=UP

+'[' UP '!=' UP ']'

+echo 'Health check status is up!'

Health check status is up!
```

PR #159, opened 2019-10-28
Add Docker Hub test hook to run Zipkin GCP with credential

saturnism:
Final piece for #130


adriancole:
maybe add a link to the docker hub build for convenience?

anuraaga:
Thanks for working on this! I forgot it's often annoying loading a service account in CI since GOOGLE_APPLICATION_CREDENTIALS must be a path to a file - maybe motivation to update google-auth libraries to fallback to JSON parsing when the environment variable doesn't resolve to a real file? ;)

Anyways, if possible we should avoid overriding the test hook and use the same docker-compose.test.yml format / command as our upstream tests.

https://github.com/openzipkin/zipkin/blob/master/docker/docker-compose.test.yml

Using a volume, I think this can be relatively simple - this compose shows a service account on my machine. Can you try using something like this?

```
version: "3.6"
volumes:
  gcp-service-account:
services:
  extract-service-account:
    image: busybox
    volumes:
      - gcp-service-account:/credentials:rw
    command: sh -c 'echo $GOOGLE_APPLICATION_CREDENTIALS_BASE64 | base64 -d > /credentials/service-account-key.json'
    environment:
      - GOOGLE_APPLICATION_CREDENTIALS_BASE64
  show-service-account:
    image: busybox
    volumes:
      - gcp-service-account:/credentials:ro
    command: sh -c 'cat /credentials/service-account-key.json'
    depends_on:
      - extract-service-account
```

saturnism:
> maybe add a link to the docker hub build for convenience?

Will do once something is running successfully; it's in a private docker hub repository atm just in case i did something wrong w/ the credentials.

saturnism:
> Thanks for working on this! I forgot it's often annoying loading a service account in CI since GOOGLE_APPLICATION_CREDENTIALS must be a path to a file - maybe motivation to update google-auth libraries to fallback to JSON parsing

I have seen quite a few use cases and requests that this would benefit :( /cc @chingor13 

> avoid overriding the test hook and use the same docker-compose.test.yml format / command as our upstream tests.

This is is great info! Will give this a try.

saturnism:
@anuraaga the docker compose hook was much easier indeed. thanks!

saturnism:
Docker test is passing now:

https://cloud.docker.com/repository/registry-1.docker.io/saturnism/zipkin-gcp/builds/a5640777-a3cc-4ad3-a0a5-e3ab72fdbde0

I feel this is ready to go.

docker/docker-compose.test.yml
```diff
@@ -0,0 +1,61 @@
...
+    volumes:
+      - gcp-service-account:/credentials:rw
+    command: sh -c 'echo $GOOGLE_APPLICATION_CREDENTIALS_BASE64 | base64 -d > /credentials/service-account-key.json'
+    environment:
+      - GOOGLE_APPLICATION_CREDENTIALS_BASE64
+  show-service-account:
+    image: busybox
+    volumes:
+      - gcp-service-account:/credentials:ro
+    command: sh -c 'ls -lh /credentials/service-account-key.json'
```
anuraaga: I almost thought we were leaking the credential ;) Probably can add `# Show that the file exists as a sanity check in logs.`

saturnism: :) yeah, i noticed the cat and changed it immediately :p

saturnism:
I've added the env var w/ the credential key into openzipkin/zipkin-gcp auto build configuration.

docker/docker-compose.test.yml
saturnism: updated w/ more comment.

PR #161, opened 2019-10-31
Update Dockerfile.release to our final conventions.

anuraaga:


adriancole:
looks good. we will have to remember to update this file when we update the pom to 2.19.0 which is just out

adriancole:
we could also read the zipkin version from maven similar to https://github.com/openzipkin/zipkin-dependencies/tree/f0608b9e5a8d3e6ce1aa96256ea8572e054a335b#running-in-a-spark-cluster

PR #162, opened 2019-11-02
Updates to latest Zipkin and Armeria

adriancole:


pom.xml
```diff
@@ -47,13 +47,13 @@
     <spring-boot.version>2.1.9.RELEASE</spring-boot.version>
     <!-- This allows you to test feature branches with jitpack -->
     <armeria.groupId>com.linecorp.armeria</armeria.groupId>
-    <armeria.version>0.94.0</armeria.version>
-    <zipkin.version>2.18.0</zipkin.version>
```
adriancole: interestingly, you'll notice we were out-of sync here with the dockerfile!

docker/Dockerfile.release
```diff
@@ -39,7 +39,7 @@ RUN export version=$(echo "${source_branch}" | cut -d '-' -f 2) && \
   (mkdir stackdriver && cd stackdriver && jar -xf ../zipkin-module-storage-stackdriver-${version}-module.jar) && \
   rm zipkin-module-storage-stackdriver-${version}-module.jar
 
-FROM openzipkin/zipkin:2.18.3
+FROM openzipkin/zipkin:2.19.1
```
adriancole: I wanted to derive this from maven version, but it seems people have already tried and failed at that... https://stackoverflow.com/questions/36556766/can-a-variable-be-used-in-docker-from

PR #169, opened 2020-02-25
Upgrade to Zipkin 2.20

anuraaga:
Verified unit tests pass with local snapshot. Less concerned about testing with real cluster since integration tests should pick up issues.


adriancole:
thanks again!

PR #171, opened 2020-02-28
Add last license updates

anuraaga:
In a strange state that local build doesn't catch these for some reason but should be fine finally.


adriancole:
this seems like the weird thing we disabled the check before on.. anyway all good to merge!

PR #172, opened 2020-04-16
updates to latest

adriancole:
verified through the test docker image


docker/Dockerfile.release
```diff
@@ -11,7 +11,6 @@
 # or implied. See the License for the specific language governing permissions and limitations under
 # the License.
 #
-# Download artifact from bintray, where the artifact should have been published by now.
```
adriancole, line 14: accident of license plugin, but incidentally it is redundant to a below comment :)

PR #176, opened 2020-05-18
Upgrades brave and updates HexCodec import to correct package

elefeint:
`spring-cloud-sleuth` is already using the latest version of Brave, but `zipkin-gcp` lagged. 

In between, `HexCodec` helper class got moved into a subpackage in openzipkin/brave#1193.

This PR will allow `spring-cloud-gcp` to upgrade to the latest `spring-cloud-sleuth` without triggering runtime class incompatibility (`Caused by: java.lang.NoClassDefFoundError: brave/internal/HexCodec`).


adriancole:
after this change I'll shade it as this was an accident. thanks!

PR #180, opened 2020-10-23
Updates to latest everything and sets minimum Java level to 8

adriancole:
JDK 15 broke retrolambda, and there's little cause for us to keep <8
compat here until we're told to. This adds rationale around that and we
can keep watch for end users who care. (Note Brave is still JRE 6+)

This also rejigs docker to use a single file for tasks.


pom.xml
```diff
@@ -36,37 +36,41 @@
...
 
     <!-- from grpc-protobuf -->
     <protobuf.version>3.12.0</protobuf.version>
     <guava.version>29.0-android</guava.version>
 
     <!-- only used for stackdriver protos, we could possibly obviate this if a problem -->
-    <grpc-google-cloud-trace.version>1.2.0</grpc-google-cloud-trace.version>
+    <grpc-google-cloud-trace.version>1.2.5</grpc-google-cloud-trace.version>
 
+
```
adriancole, line 70:
```suggestion
```

anuraaga (approved):

PR #181, opened 2020-10-25
Moves build to Sonatype and GitHub Container Repository publishing 

adriancole:
This makes publication the same as Zipkin, which no longer uses bintray,
and primarily references docker from ghcr.io.

---

Before, we intermediated publications through Bintray. Over the years,
we have had difficulty with this including timeouts to Maven central,
extra administration, service outages and quota problems.

Currently, the OpenZipkin org cannot release anything because something
outside our control is downloading zipkin-server too frequently. While
I've contacted Bintray support on this, the advice returned was to pay
for a premium edition to block the IPs. Rather than get into deciding
who can download Zipkin or not, this change takes out the intermediary.

Instead, we go to "old school" meaning we publish directly to Sonatype,
which was already a service dependency of ours anyway. For example,
even using Bintray, we had to encrypt and supply Sonatype credentials
so that binaries could get to Maven central. The main difference here is
the GPG signature process automated by Bintray is now self-managed.

What this implies is we have to encrypt a signing key here, which I've
done (and also used for brave-karaf). If a problem happens with that
key, for example expiration, a new one would need to be encrypted into
the CI variables. We would also need to update the quickstart to allow
that key to not trigger scary messages:

https://github.com/openzipkin/openzipkin.github.io/blob/master/quickstart.sh

The changes here were editing files, and some commands needed in order
to encrypt variables. The non-obvious ones are mentioned here:
```
travis encrypt GPG_PASSPHRASE=my_passphrase
gpg --export-secret-keys > travis/signing_key.gpg
travis encrypt-file travis/signing_key.gpg
rm travis/signing_key.gpg
mv signing_key.gpg.enc travis/signing_key.gpg.enc
```

This process was tested on brave-karaf up to and including a release.
It is possible, there's a slight glitch, though, so any notes post
merge will end up here.


adriancole:
cc @openzipkin/core @openzipkin/devops-tooling you might be aware that we cannot do anything right now to being over download quota. For example, we can't even manually deploy using Bintray. This automates how I was manually deploying zipkin the last few days.

adriancole:
PS this isn't the first time we've been over quota, just the first time it has happened for a reason beyond our control. For example, we've had some "dev zero" incidents in the past where @abesto had to delete large ranges of versions so we can deploy again. the difference here is we cannot control who downloads zipkin. It is basically the straw that breaks the camel's back and manually deploying is something I wish on no one.

adriancole:
one thing I'd like some input on is the key itself. I made it with my name and email but that's probably not good and we can change it.

For example, we can use "zipkinci+zipkin-dev@googlegroups.com" as we do in other places or "zipkin-admin@googlegroups.com" as we do for things like our DNS stuff. I'd like some input from other committers also on the duration of the key as I just made a 1 year expiring one.

abesto:
Ohai! My 2c (no real skin in the game so grain of salt, but maybe useful):
* High level, the goal and approach make sense to me
* zipkin-admin@ looks like a good bet. In case e-mail verification or whatever is needed in the future, it's good to use a restricted-but-not-single-person email group.
* 1 year for the key expiry is sane I think. I assume we'd still distribute the public key via the usual public key servers, which means if needed, a revocation can be published, so the key can be rotated more frequently in case of compromise. Might be a good idea to add some loud warning (maybe set up a dedicated CI job?) to renew the key about a month in advance.

adriancole:
@abesto great idea. We already have a job on openzipkin.github.io repo to scream about DNS.

ps notes include uploading key to ubuntu and updating quickstart so that people aren't scared :D

anuraaga (approved):
Is there much advantage to a one-year expiration? It always seems to me that a one year key has the tedium of rotation every year, but the same downside as longer keys in that if it leaks, then someone can go wild for up to a year unless it's revoked.

adriancole:
It seems the key we were using from Bintray had no expiration

```
gpg --list-keys
/Users/adriancole/.gnupg/pubring.kbx
------------------------------------
pub   rsa3072 2020-10-23 [SC] [expires: 2022-10-23]
      3FAD387B57E8FE68E5FA632A9F79E8538DCFC895
uid           [ unknown] Adrian Cole <adrian.f.cole@gmail.com>
sub   rsa3072 2020-10-23 [E] [expires: 2022-10-23]

pub   rsa4096 2015-02-17 [SC]
      8756C4F765C9AC3CB6B85D62379CE192D401AB61
uid           [ unknown] Bintray (by JFrog) <bintray@bintray.com>
sub   rsa4096 2015-02-17 [E]
```

In this case, we've used it for 5 years :) Maybe we can say 5 year expiration based on this precedent @abesto :D

It has proved a pain for sure to update things like DNS annually, especially on volunteers..

adriancole:
FYI We received this today from bintray:

> I have worked with our internal team and further increased the download quota limit to 5TB. We have also unblocked the "openzipkin" organization.

If this changes any ones opinion on doing this change, please comment back. Honestly I don't know if 5TB will accommodate the people who over-downloaded and caused our org to lock. Gut feel is that while this is a nice gesture and much appreciated, at lease our main org (ex openzipkin not necessarily openzipkin-contrib) should probably go directly to sonatype so that we don't need to break glass when something like this happens again.

anuraaga:
@adriancole Any idea how much the quota was before? Straight to sonatype seems fine in principle, but one nice reason to stay with bintray is not having to go through all these motions in all the repos which is some work.

adriancole:
If we want a no hassle thing, there's no difference between making a non-expiring key and an api key that is never changed is there? There is actually one less credential to manage :D

As the person whose dealt with the manual impact of the last 5 years I can say I like the idea of going directly to sonatype. timeouts pushing to bintray have cost me many evenings.

adriancole:
in other words apples to apples is we use a non-expiring key. not expiring key vs bintray's non-expiring one.

adriancole:
The other thing I should mention that most folks don't know is that provisioning a new project is zero work with sonatype, but it has the following with bintray:

1. create a new repo in bintray (add description, github url etc), click publish snapshots
2. setup the same name in the pom as above
3. publish a build manually as you can't add it to jcenter until you do one manual build
4. click and make sure you add the correct group Id that matches the pom
5. wait for the email to come back that jcenter is approved
6. manually release that in bintray and click the sonatype thing to sync to central

In not using bintray, there's none of the above steps. if your sonatype has access to io.zipkin, you can setup endless projects with no admin overhead except this particular piece.

my 2p

adriancole:
one downside vs status quo is that bintray when up (no 503 errors etc which have happened several times), we get an immediately synced archive repo which we can build docker image from without waiting for maven to sync.

On the other hand it is fairly trivial to re-use the binary we just pushed in Docker on release, and I'd prefer to do that. For example, pushing the tag explicitly docker-N.N.N could take from central, but when we literally just published the jars, we could build a docker image from that local copy.

anuraaga:
Ah by work I wasn't referring to key - just having to do what we did here to all our repos. So in other words up to you on if you want to do it :P

adriancole:
yep I'm assuming whatever ground level infra action, I'll do that. For me, I like not having to worry about something when I have no time, and I have some now heh

adriancole:
After doing some thinking about why Bintray doesn't set an expiry, I researched a little and it seems that there's a solid case for not ever expiring a key as it is a false sense of security https://security.stackexchange.com/questions/14718/does-openpgp-key-expiration-add-to-security/79386#79386

Instead, we can set expires to zero and privately share the revocation certificate among core committers (as this key will be most conveniently used by the org, it should be owned by @openzipkin/core)

need a couple thumbs up on this.

adriancole:
https://github.com/openzipkin/zipkin/pull/3275 here's the upstream in zipkin which is finally ready!

adriancole:
@saturnism I've mostly setup github action to verify the docker image process. This combines what we did in travis and docker hub before into one flow. It is currently broken on missing credentials. I'm going to leave this red as it is a TODO. If you have time, please kick it into gear via https://github.com/openzipkin/zipkin-gcp/issues/182

.github/workflows/docker-tests.yml
```diff
@@ -0,0 +1,94 @@
...
+        uses: actions/cache@v1
+        with:
+          path: ~/.npm
+          key: npm-packages-${{ runner.os }}-${{ hashFiles('zipkin-lens/package-lock.json') }}
+      - name: Cache Maven Modules
+        uses: actions/cache@v1
+        with:
+          path: ~/.m2/repository
+          key: m2-repository-${{ hashFiles('**/pom.xml') }}
+      - name: Cache Maven Wrapper
```
anuraaga, line 38: Not related to this PR just observation since I only realized it now, I wonder if it's idiomatic to check the wrapper into the repo?

https://github.com/takari/maven-wrapper/tree/master/.mvn/wrapper

anuraaga (approved):

.github/workflows/docker-tests.yml
adriancole: we did then we reverted in in our apache experiment as there was a fuss about binaries in git. we could have renamed it to jpg to get around that :D

meanwhile I suppose we could do that as it is simpler than all these little stages, and rarely changes.

adriancole: nuked it here and in zipkin. the topic is indeed distracting.. there are source files larger than that binary, I'd bet :P

adriancole:
yay fails for the right reason

```
org.springframework.beans.factory.BeanCreationException: Error creating bean with name 'googleCredentials' defined in zipkin.module.storage.stackdriver.ZipkinStackdriverStorageModule: Bean instantiation via factory method failed; nested exception is org.springframework.beans.BeanInstantiationException: Failed to instantiate [com.google.auth.Credentials]: Factory method 'googleCredentials' threw exception; nested exception is java.lang.IllegalArgumentException: no JSON input found
```

PR #187, opened 2020-11-10
Updated credentials verbiage for GitHub Actions

saturnism:
Follow up on https://github.com/openzipkin/zipkin-gcp/issues/182#issuecomment-724445747

The verbiage is only limited to GCP Service Account configuration. There are additional Travis-mentions for releasing etc - those verbiages are not updated in this PR.


adriancole (approved):
cool!

odd the build is red.. I wonder if we have to turn off pull requests for this action?

PR #195, opened 2021-12-21
Update Zipkin to address CVE-2021-44228

DannyPat44:
Updated the Zipkin version to address [#Issue-194](https://github.com/openzipkin/zipkin-gcp/issues/194)


shakuzen (commented):
I think the license plugin wants the license headers to be updated to end in the current year.

shakuzen (approved):

PR #212, opened 2024-01-13
ci: checks markdown for bad links

codefromthecrypt:
want to automate this so we don't have users complain into the void https://github.com/openzipkin/zipkin-gcp/issues/198


codefromthecrypt:
ok if folks are keen on this, I can propagate to the other repos same way.

reta (approved):

.github/workflows/test_markdown.yml
```diff
@@ -0,0 +1,33 @@
+# yamllint --format github .github/workflows/test_markdown.yml
+---
+name: test_markdown
+
+on:  # yamllint disable-line rule:truthy
```
codefromthecrypt, line 5: TIL why on was making truthy warnings.. here are some details @openzipkin/devops-tooling https://github.com/adrienverge/yamllint/issues/430

now, finally we have one workflow that has no lint output

PR #213, opened 2024-01-14
ci: ensures markdown config doesn't trigger java tests

codefromthecrypt:
This also fixes lint on the test workflow


reta (approved):

codefromthecrypt:
replaced with same setup as zipkin, so they use exactly same lint now

PR #217, opened 2024-02-18
removes invalid openssl check and bumps deps

codefromthecrypt:
@provegard mentioned on gitter that we are getting an openssl error. I think the check is no longer valid after we switched to latest JRE + spring boot 3, as when I remove the check, the docker image (alpine JRE 21) works fine. So, I'm removing the check and in worst case, if someone has a misconfiguration, they get a late error in /health, but we don't get a false report for everybody else.

Traces created with https://github.com/openzipkin/brave-example/tree/master/armeria and self-tracing too.
I ran the locally build docker image like so, configured with `Trace Agent` role

```bash
$ docker run --rm -p 9411:9411 --name zipkin-gcp -e SELF_TRACING_ENABLED=true  -e STORAGE_TYPE=stackdriver   -e GOOGLE_APPLICATION_CREDENTIALS=/zipkin/.gcp/credentials.json   -e STACKDRIVER_PROJECT_ID=adrian-test-414623    -v $HOME/.gcp:/zipkin/.gcp:ro   openzipkin/zipkin-gcp:test
```

Screenshot 2024-02-18 at 08 15 24

### DIFF
--- a/build-bin/docker-compose-zipkin-gcp.yml
+++ b/build-bin/docker-compose-zipkin-gcp.yml
@@ -3,14 +3,14 @@ volumes:
   gcp-service-account:
 services:
   extract-service-account:
-    image: ghcr.io/openzipkin/alpine:3.19.0
+    image: ghcr.io/openzipkin/alpine:3.19.1
     volumes:
       - gcp-service-account:/credentials:rw
     command: -c 'echo $GOOGLE_APPLICATION_CREDENTIALS_BASE64 | base64 -d > /credentials/service-account-key.json'
     environment:
       - GOOGLE_APPLICATION_CREDENTIALS_BASE64
   show-service-account:
-    image: ghcr.io/openzipkin/alpine:3.19.0
+    image: ghcr.io/openzipkin/alpine:3.19.1
     volumes:
       - gcp-service-account:/credentials:ro
     # Show that the file exists as a sanity check in logs.
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -13,7 +13,7 @@
 #
 
 # zipkin version should match zipkin.version in /pom.xml
-ARG zipkin_version=3.0.5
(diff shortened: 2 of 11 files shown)


codefromthecrypt:
ps we once had a test account and maybe I accidentally stopped it from running when updating CI config and/or removing travis.. I opened https://github.com/openzipkin/zipkin-gcp/issues/218 to get automated integration test with a real account working again, which would have prevented not noticing this. I ad-hoc tested the server once in the last couple months, but broke it after!

reta (approved):

codefromthecrypt:
I found when the docker CI workflow was accidentally deleted.. several years ago. running the tests again in case the credentials still work.

codefromthecrypt:
ok rolling back CI as the google-cloud account was revoked, which is not terribly surprising the years

```
{
  "status" : "DOWN",
  "zipkin" : {
    "status" : "DOWN",
    "details" : {
      "StackdriverStorage{zipkin-gcp-ci}" : {
        "status" : "DOWN",
        "details" : {
          "error" : "com.google.auth.oauth2.GoogleAuthException: Error getting access token for service account: 400 Bad Request\nPOST [https://oauth2.googleapis.com/token\n{\](https://oauth2.googleapis.com/token/n%7B/)"error\":\"invalid_grant\",\"error_description\":\"Invalid grant: account not found\"}, iss: zipkin-gcp-ci-github@zipkin-gcp-ci.iam.gserviceaccount.com"
        }
      }
    }
  }
}
```

we'll just do ad-hoc integration tests until someone at google gives us an account again.

codefromthecrypt:
thanks for the look @reta!

codefromthecrypt:
https://github.com/openzipkin/zipkin-gcp/releases/tag/2.2.2

provegard:
I just tested the 2.2.2 tag, and it works like a charm. Thank you for the fix!

PR #235, opened 2025-02-17
updates docker and library dependencies to the latest version

codefromthecrypt:
This is me playing with goose again. I'm aware that much of this could be done with openrewrite, but I am intentionally trying to practice prompting and learn more. Me doing this for experience doesn't in any way say this approach is more appropriate than other tools like openrewrite or dep bumpers. Exploring tech is the fun part of volunteer work!

Anyway... Here's what I did:

* installed [goose](https://block.github.io/goose/docs/getting-started/installation)
* configured goose with google model: gemini-2.0-flash as it is [free](https://ai.google.dev/pricing#2_0flash)
* added [maven-mcp-server](https://github.com/Bigsy/maven-mcp-server) so that it can see latest versions without using RAG 
* added a preparatory commit for some things easier to do than explain to a prompt, notably docker image updates and some relationship between them and shell scripts as well as a recent break in config by errorprone
* ran the following prompt in goose, noticed it didn't see zipkin's last update (probably not in the central update), so updated it. I also ran goose again with the same prompt as I noticed a couple versions didn't bump

If you notice any misses, lemme know and/or update them!

## Details

Here's the extension I added to ~/.config/goose/config.yaml for [maven-mcp-server](https://github.com/Bigsy/maven-mcp-server)
```yaml
  maven:
    args:
    - mcp-maven-deps
    cmd: npx
    enabled: true
    envs: {}
    name: maven
    type: stdio
```

Here's the goose prompt. I didn't change it much from zipkin-reporter so it includes some instructions not necessary here, but I was too lazy to remove them.
```
Update all dependency and plugin versions to latest in all pom.xml files, including child modules and bom, including those set with properties whose name end in ".version".

Here are the rules:
* If the latest version includes a word like beta, alpha or rc, use the latest version that doesn't.
* Do not update any versions for this project's groupId
* Do not replace a property expansion like ${netty.version} for an explicit version. Instead, update the property value.
* If a version value begins and ends in '@', it is set by a property with a name equal to the text between the '@' characters.
* If a dependency sets version by property, update the property value, don't expand it.
* If there is a comment above its definition, honor any constraint it might have. Do not update versions with a TODO comment
* If the version property name contains digits before ".version", do not upgrade past that major version. For example, the "spring5.version" means update to the latest version of spring 5.x.x.
* If the version property name starts with "floor-", update to the latest patch version, but not to the next minor or major.

Rules when making file modifications
* keep the same indentation, formatting as the original files
* Do not change the order of properties you affect

Once complete, test your change by running `./build-bin/test`.

```

sometimes I got rate-limited and just entered "continue" after a few moments


reta (approved):

codefromthecrypt:
cc @making fyi since armeria is out-of-date with netty, I added a commit that would need to happen in zipkin-otel as well (to use the bom to align the version. remember we don't do that in the parent pom, only in modules where armeria is a dep)

